PR #55, opened 2022-03-15
fix(schema): fix the typehint for nested extractors

Lee-000:


GravitiRob:
Task linked: [CU-2cy5uvx 确定 field name + type 到 extractor + dtype 映射的格式，并实现 filename 作为 example](https://app.clickup.com/t/2cy5uvx)

graviti/schema/extractors.py
```diff
@@ -17,7 +17,16 @@
 _T = TypeVar("_T")
 _D = Dict[str, Tuple[Iterator[_T], str]]
 _Extractor = Tuple[Callable[[Dict[str, Any]], Iterator[_T]], str]
-_Extractors = Dict[str, _Extractor[_T]]
+
+
+class Extractors(Protocol, Generic[_T]):
```
linjiX, line 22: Why we need `Generic`?

linjiX: Will a private class better?

linjiX (changes requested):

graviti/schema/extractors.py
graczhual: > Will a private class better?

This class will be used in lazy.py

linjiX: Other types around it are all private?

linjiX: > > Why we need `Generic`?
> 
> To specify the type of the values in `Iterator`.

But all _T in `Extractors` is `Any`?

graviti/schema/extractors.py
```diff
@@ -17,7 +17,16 @@
 _T = TypeVar("_T")
 _D = Dict[str, Tuple[Iterator[_T], str]]
 _Extractor = Tuple[Callable[[Dict[str, Any]], Iterator[_T]], str]
-_Extractors = Dict[str, _Extractor[_T]]
+
```
linjiX, line 20: Why not just set `Any` in `_Extractor`

graczhual (approved):

PR #63, opened 2022-03-24
feat(schema): implement the schema imports system

linjiX:


GravitiRob:
Task linked: [CU-2dn09j3 实现 Portex imports 语法](https://app.clickup.com/t/2dn09j3)

graviti/schema/package.py
```diff
@@ -287,12 +320,52 @@ def __getitem__(self, key: str) -> Type["PortexType"]:
...
+        subpackage[key] = value
+
+    def __iter__(self) -> Iterator[str]:
+        return chain(*self._subpackages.values())
+
+    def update(self, other: "Imports") -> None:
+        """Update the imports with another imports.
+
+        Arguments:
+            other: A imports whose types need to be updated to this imports.
```
Lee-000, line 352: `A imports`?

graviti/schema/param.py
```diff
@@ -131,6 +132,45 @@ def check(self, arg: Any) -> Any:
 
         return arg
 
+    def load(self, content: Any, imports: Imports) -> Any:
+        """Create a instance of the parameter type from the python content.
```
Lee-000, line 136: `an instance`

graviti/schema/field.py
```diff
@@ -16,8 +19,43 @@ class Field(NamedTuple):
...
+
+        """
+        return cls(content["name"], PortexType.from_pyobj(content, imports))
+
+    def to_pyobj(self) -> Dict[str, Any]:
+        """Dump the instance to a python dict.
 
-class Fields(Sequence[Field]):
+        Returns:
+            A Python dict representation of the field.
```
Lee-000, line 52: `A ... representation` does not seem right.

Lee-000 (changes requested):

graviti/schema/field.py
```diff
@@ -16,8 +19,43 @@ class Field(NamedTuple):
...
+    def imports(self) -> Imports:
+        """Property imports to indicate the imports of this field.
+
+        Returns:
+            The imports of this field.
+
+        """
+        imports = Imports()
+        imports.update_from_type(self.type)
+        return imports
```
Lee-000, line 32: Why create a new `Imports` instance each time to get the `imports` property?

graviti/schema/package.py
```diff
@@ -141,6 +154,21 @@ def __init__(self, package: ExternalPackage) -> None:
...
+        type_ = self._data.get(key)
+        if type_:
+            if type_ != value:
+                raise KeyError("Duplicate names")
+            return
+
+        for name, type_ in self._data.items():
+            if type_ == value:
+                if name != key:
+                    raise ValueError("Same type with the different names in subpackage")
```
Lee-000, line 167: Would "Same type with a different name already exists in the subpackage" be more clear?

graviti/schema/package.py
```diff
@@ -287,12 +320,52 @@ def __getitem__(self, key: str) -> Type["PortexType"]:
...
+        return chain(*self._subpackages.values())
+
+    def update(self, other: "Imports") -> None:
+        """Update the imports with another imports.
+
+        Arguments:
+            other: An :class:`Imports` instance whose types need to be updated to this imports.
+
+        """
+        assert not self._package and not other._package  # pylint: disable=protected-access
```
Lee-000, line 355: Why can't they have packages?

graviti/schema/ptype.py
```diff
@@ -33,6 +34,33 @@ def check(_: TypingAny) -> TypingAny:
         """
         raise NotImplementedError
 
+    @staticmethod
+    def load(content: TypingAny, _: Imports) -> TypingAny:
+        """Create a instance of the parameter type from the python content.
```
Lee-000, line 39: `An instance`

graviti/schema/ptype.py
```diff
@@ -33,6 +34,33 @@ def check(_: TypingAny) -> TypingAny:
...
+
+        Arguments:
+            content: A python presentation of the parameter type.
+            _: The imports of the parameter type.
+
+        Returns:
+            A instance of the parameter type.
```
Lee-000, line 46: `An instance`

Lee-000 (changes requested):

graviti/schema/field.py
linjiX: Because Field is a `NamedTuple` we need to keep it the same behavior with a normal tuple (name, type)

graviti/schema/package.py
linjiX: the base package in import is used for building extending package and local package, normally the these two process should not call this `update` method. I just add a simple check here to avoid unexpected error. I will remove the base class from Imports and use another class for package building in the future.

Lee-000 (approved):

PR #70, opened 2022-03-25
feat(schema): support accessing Portex types in package by attr-style

linjiX:


GravitiRob:
Task linked: [CU-2dn0caa 支持通过 "." 方法获取 package 中的类型](https://app.clickup.com/t/2dn0caa)

graviti/utility/attr.py
```diff
@@ -0,0 +1,141 @@
...
+    overload,
+)
+
+from tensorbay.utility import ReprMixin, ReprType
+
+_T = TypeVar("_T")
+_D = TypeVar("_D")
+
+
+class _AttrDict(Generic[_T], ReprMixin):
```
Lee-000, line 27: Should the order of the methods follow the standard?

graviti/utility/attr.py
```diff
@@ -0,0 +1,141 @@
...
+            return self._data.__getitem__(key)
+
+        return self._modules.__getitem__(prefix).__getitem__(suffix)
+
+    def _repr_head(self) -> str:
+        return AttrDict.__name__
+
+
+class AttrDict(_AttrDict[_T], Mapping[str, _T]):
+    """A dict which allows for attr-style access of values."""
```
Lee-000, line 114: `A dict that allows attr-style access of values.`?

graviti/utility/attr.py
```diff
@@ -0,0 +1,141 @@
...
+            self._getitem(key)
+        except KeyError:
+            return False
+
+        return True
+
+    def __getitem__(self, key: str) -> _T:
+        try:
+            return self._getitem(key)
+        except KeyError:
```
Lee-000, line 59: Would there be `AttributeError`?

graviti/utility/attr.py
```diff
@@ -0,0 +1,141 @@
...
+            module = _AttrDict()
+            self._modules.__setitem__(prefix, module)
+
+        module.__setitem__(suffix, value)
+
+    def __getattr__(self, key: str) -> Union["_AttrDict[_T]", _T]:
+        try:
+            return self._data.__getitem__(key)
+        except KeyError:
+            pass
```
Lee-000, line 86: Which one would be better, `try except` or `__contains__`?

graviti/utility/attr.py
```diff
@@ -0,0 +1,141 @@
...
+            return self._getitem(key)
+        except KeyError:
+            return self.__missing__(key)
+
+    def __setitem__(self, key: str, value: _T) -> None:
+        try:
+            prefix, suffix = key.split(".", 1)
+        except ValueError:
+            if key in self._modules:
+                raise KeyError("Key prefix conflict") from None
```
Lee-000, line 67: Setting an entire module is not allowed, but setting an item in one module is accepted?

Lee-000 (changes requested):

graviti/utility/attr.py
linjiX: `__getitem__` should raise `KeyError`

graviti/utility/attr.py
linjiX: copyed from another python library `AttrDict`
https://github.com/bcj/AttrDict/blob/8c1883162178a124ee29144ca7abcd83cbd9d222/attrdict/mixins.py#L159

Lee-000: Use `that` instead of `which`?

graviti/utility/attr.py
linjiX: `__contains__` is faster when accessing non-leaf node
`try except` is faseter when getting leaf node

graviti/utility/attr.py
linjiX: `AttrDict` does not provide the ability to set non-leaf node manually,
`__setitem__` will treat all values input from user as leaf node even it is another `AttrDict`

Lee-000 (approved):

PR #94, opened 2022-04-13
refactor: use imports from modules instead of imports from files

Lee-000:
- Use imports from modules instead of imports from files
when importing from other modules.
- Move `schema/pyarrow.py` to `utility/pyarrow.py`.
- Add necessary imports in `__init__.py` in multiple modules.


graczhual (approved):

graviti/client/branch.py
```diff
@@ -8,7 +8,7 @@
 from typing import Any, Dict, Optional
 from urllib.parse import urljoin
 
-from graviti.utility.request import URL_PATH_PREFIX, open_api_do
+from graviti.utility import URL_PATH_PREFIX, open_api_do
```
AChenQ, line 11: Do we need to split it into several commits？

graviti/utility/__init__.py
```diff
@@ -5,17 +5,37 @@
 """Utility module."""
 
 from graviti.utility.attr import AttrDict
-from graviti.utility.file import File
-from graviti.utility.lazy import LazyFactory
+from graviti.utility.file import READ_ONLY_URL, File
+from graviti.utility.lazy import LazyFactory, LazyList
+from graviti.utility.pyarrow import (
```
linjiX, line 10: why we need to move `pyarrow` into `utility`?

linjiX (approved):

AChenQ (approved):

PR #95, opened 2022-04-14
refactor(utility): implement File without inheritance from Tensorbay

Joey9933:


GravitiRob:
Task linked: [CU-2jn2ujq TensorBay 代码迁移：重写graviti.utility.File 类](https://app.clickup.com/t/2jn2ujq)

graviti/utility/file.py
```diff
@@ -25,16 +76,61 @@ class File(RemoteFileMixin):
     __slots__ = ("url", "_checksum")
 
     def __init__(self, url: str, checksum: str) -> None:
-        super().__init__("")
         self.url = READ_ONLY_URL(url)
         self._checksum = checksum
 
     def _repr_head(self) -> str:
-        return f'{self.__class__.__name__}("{shorten(self._checksum)}")'
+        return f'{self.__class__.__name__}("{self._checksum[:7]}")'
```
linjiX, line 83: why stoping using `shorten`?

graviti/utility/file.py
```diff
@@ -6,14 +6,65 @@
...
+from typing import Callable, Optional, Union
+
+from _io import BufferedReader
+from tensorbay.utility.repr import ReprMixin
+
+from graviti.exception import ResponseError
+from graviti.utility.requests import UserResponse, config, get_session
+
+
+class URL:
```
linjiX, line 18: seems we do not need `URL` class for now, can we just replace it with `str`

graviti/dataset/dataset.py
```diff
@@ -9,13 +9,14 @@
 
 from tensorbay.dataset import Notes, RemoteData
 from tensorbay.label import Catalog
+from tensorbay.utility import URL
```
linjiX, line 12: your commit message is not clear.
try to directly describe what have you done,
like `implement xxx` or `add xxx`

linjiX (changes requested):

graviti/utility/file.py
Joey9933: the goal of `shorten` is easy to achieve, why should we use that simple function...

Lee-000 (changes requested):

graviti/utility/file.py
Joey9933: for now, we need `URL` only in file.py, by `READ_ONLY_URL`, which is needed by `File`  class

Joey9933: just use `url:str`

graviti/utility/file.py
```diff
@@ -25,16 +26,50 @@ class File(RemoteFileMixin):
...
-        super().__init__("")
-        self.url = READ_ONLY_URL(url)
+        self.url = url
         self._checksum = checksum
 
     def _repr_head(self) -> str:
         return f'{self.__class__.__name__}("{shorten(self._checksum)}")'
 
+    def _urlopen(self) -> UserResponse:
+        if not self.url:
```
linjiX, line 36: In what condition `self.url` does not exist?

graczhual (approved):

graviti/utility/file.py
```diff
@@ -25,16 +26,50 @@ class File(RemoteFileMixin):
...
+            raise ValueError(f"The file cannot open because {self._repr_head()} has no url")
+
+        session = get_session()
+        return UserResponse(
+            session.request("GET", self.get_url(), timeout=config.timeout, stream=True)
+        )
+
+    def get_url(self) -> str:
```
linjiX, line 44: if self.url always exists, shall we still need `get_url` method?

linjiX (changes requested):

graviti/utility/file.py
```diff
@@ -25,16 +26,50 @@ class File(RemoteFileMixin):
...
+        """
+        if not self.url:
+            raise ValueError(f"The file URL cannot be got because {self._repr_head()} has no url")
+
+        return self.url
+
+    def open(self) -> Union[UserResponse, BufferedReader]:
+        """Return the binary file pointer of this file.
+
+        The remote file pointer will be obtained by ``urllib.request.urlopen()``.
```
AChenQ, line 62: We do not use `urlopen()` now, we can remove this line.

AChenQ (changes requested):

graviti/utility/file.py
Joey9933: ok

graviti/utility/file.py
Joey9933: if `self.url` is a null string, this line will raise the error.

linjiX: Is that a normal case when url is an empty string?
If not, we can just remove the check.

Joey9933: this check works for illegal null string url. in terms of robustness, I prefer remaining this check

linjiX: If we add the check here, which means we need to add the check for all the string in the code.
Also empty string is an illegal case, wrong url, wrong type are also illegal, that means we need check them all.

If in normal case it will not be empty, there is no need to check it.

Joey9933: assuming we remain this check here, the reported error could let us know that the url is null; if the url is not null but illegal, it will cause error when we are trying to open the file. I think this is clearer to distinguish errors.

linjiX: Empty string will also trigger exception when open the file.

Joey9933: that's true. but is the exception info caused by null string the same as that of other wrong url strings?

linjiX: You can try it.

graviti/utility/file.py
```diff
@@ -25,16 +26,28 @@ class File(RemoteFileMixin):
...
 
+    def _urlopen(self) -> UserResponse:
+        session = get_session()
+        return UserResponse(session.request("GET", self.url, timeout=config.timeout, stream=True))
+
+    def open(self) -> Union[UserResponse, BufferedReader]:
+        """Return the binary file pointer of this file.
+
+        Returns:
+            The remote file pointer for this data.
```
Lee-000, line 43: Is there a `data` concept here?

linjiX (approved):

AChenQ (approved):

Lee-000 (approved):

PR #96, opened 2022-04-14
refactor: merge folders "manager" and "dataset"

graczhual:
Dataset and draft/commit are not on the same level. DatasetManager and
DraftManager are also not on the same level. To avoid the problem of
circular import, merge them together.


graviti/manager/commit.py
```diff
@@ -47,7 +131,7 @@ def _generate(
         )
 
         for item in response["commits"]:
-            yield Commit.loads(item)
+            yield Commit.from_pyobj(item)
```
linjiX, line 134: should this be done in this task?
https://app.clickup.com/t/2k67b5k

linjiX (approved):

AChenQ (approved):

PR #98, opened 2022-04-15
feat(manager): add class "Branch" to support basic version control

graczhual:


GravitiRob:
Task linked: [CU-2k67axm 版本管理基础类：Branch](https://app.clickup.com/t/2k67axm)

graviti/manager/branch.py
```diff
@@ -3,19 +3,59 @@
...
+
+    """
+
+    parent_commit_id: str = _attr()
+    title: str = _attr()
+    description: str = _attr()
+    committer: str = _attr()
+    committed_at: str = attr()
+
+    def _loads(self, contents: Dict[str, str]) -> None:
```
linjiX, line 46: We should not use `load` anymore, @Lee-000 will help on removing the loads and dumps in the future.

graviti/manager/branch.py
```diff
@@ -3,19 +3,59 @@
...
+        commit_id: The commit id.
+        parent_commit_id: The parent commit id.
+        title: The commit title.
+        description: The commit description.
+        committer: The commit user.
+        committed_at: The time when the draft is committed.
+
+    """
+
+    parent_commit_id: str = _attr()
```
linjiX, line 40: why these attrs are dynamic?

linjiX: Also why not just inherit these members from `NamedCommit`？

graviti/manager/branch.py
graczhual: Because this _load is associated with common_load, we need another pr to modify them.

graviti/manager/branch.py
graczhual: This is the same design as before. When we create a dataset, and get branch main, we will get a response like this:
```
{'name': 'main', 'commitId': '00000000000000000000000000000000', 'parentCommitId': '00000000000000000000000000000000', 'title': '', 'description': '', 'committer': None}
```
Here, we do not need to print title, description and so on. Considering that the new design scheme has not been decided yet, should we use the same design as before or just inherit these members from NamedCommit?

graviti/manager/branch.py
```diff
@@ -3,19 +3,59 @@
...
+    title: str = _attr()
+    description: str = _attr()
+    committer: str = _attr()
+    committed_at: str = attr()
```
linjiX, line 44: should `commited_at` be dynamic?

linjiX (changes requested):

linjiX (approved):

PR #100, opened 2022-04-15
feat(manager): migrate struct related source code from TensorBaySDK

wangyuqing0424:


GravitiRob:
Task linked: [CU-2jn2tye TensorBay 代码迁移：client/struct.py](https://app.clickup.com/t/2jn2tye)

wangyuqing0424:
We will write new struct.py instead of the old one in tensorbay

PR #102, opened 2022-04-15
refactor(utility): import ReprMixin from graviti instead of tensorbay

Joey9933:


graviti/utility/file.py
```diff
@@ -9,8 +9,8 @@
 
 from _io import BufferedReader
 from tensorbay.cli.utility import shorten
-from tensorbay.utility.repr import ReprMixin
```
linjiX, line 12: import is a verb
"import repr from graviti instead of tensorbay" is OK

linjiX: Also the object we import is `ReprMixin` not `repr`

linjiX (approved):

Lee-000 (approved):

PR #106, opened 2022-04-19
refactor(schema): change the parent class of Fields to MutableMapping

Lee-000:


GravitiRob:
Task linked: [CU-2840ymr 修改 Fields 父类为 MutableMapping](https://app.clickup.com/t/2840ymr)

graviti/utility/user.py
```diff
@@ -355,3 +357,145 @@ def update(self, __m: Any = (), **kwargs: _V) -> None:  # pylint: disable=argume
 
         """
         self._data.update(__m, **kwargs)
+
+
+class OrderedDict(MutableMapping[str, _V], UserMapping[str, _V]):
```
linjiX, line 362: `UserOrderedDict`？

linjiX: If you need to rewrite the `__getitem__` method, you should not set `UserMapping` as base class

linjiX (changes requested):

graviti/utility/user.py
```diff
@@ -355,3 +357,145 @@ def update(self, __m: Any = (), **kwargs: _V) -> None:  # pylint: disable=argume
...
+            self._keys.append(key)
+            self._data[key] = value
+
+    def __getitem__(self, key: Union[int, str]) -> _V:
+        if isinstance(key, int):
+            key = self._keys.__getitem__(key)
+
+        return self._data.__getitem__(key)
+
+    def __setitem__(self, key: str, value: _V) -> None:
```
linjiX, line 387: shall we support int key?

graviti/utility/user.py
```diff
@@ -355,3 +357,145 @@ def update(self, __m: Any = (), **kwargs: _V) -> None:  # pylint: disable=argume
...
+        if not self._data.__contains__(key):
+            self._keys.append(key)
+
+        self._data.__setitem__(key, value)
+
+    def __delitem__(self, key: Union[int, str]) -> None:
+        if isinstance(key, int):
+            self._keys.__delitem__(key)
+            key = self._keys.__getitem__(key)
```
linjiX, line 396: get after delete?
why not try list.pop?

graviti/utility/user.py
```diff
@@ -355,3 +357,145 @@ def update(self, __m: Any = (), **kwargs: _V) -> None:  # pylint: disable=argume
...
+        return self._data.__contains__(key)
+
+    def __iter__(self) -> Iterator[str]:
+        return self._keys.__iter__()
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, self.__class__):
+            return False
+
+        return self._data.__eq__(other._data) and self._keys.__eq__(other._keys)
```
linjiX, line 412: shall we check `keys` first?
will it be faster?

graviti/utility/user.py
```diff
@@ -355,3 +357,145 @@ def update(self, __m: Any = (), **kwargs: _V) -> None:  # pylint: disable=argume
...
+    def keys(self) -> Tuple[str, ...]:  # type: ignore[override]
+        """Get all element keys in order.
+
+        Returns:
+            A tuple containing all elements keys.
+
+        """
+        return tuple(self._keys)
+
+    def items(self) -> Tuple[Tuple[str, _V], ...]:  # type: ignore[override]
```
linjiX, line 423: shall we return an `Iterator`?

graviti/utility/user.py
```diff
@@ -355,3 +357,145 @@ def update(self, __m: Any = (), **kwargs: _V) -> None:  # pylint: disable=argume
...
+
+    @overload
+    def pop(self, key: str) -> _V:  # pylint: disable=arguments-differ
+        ...
+
+    @overload
+    def pop(self, key: str, default: Union[_V, _T] = ...) -> Union[_V, _T]:
+        ...
+
+    def pop(self, key: str, default: Any = __marker) -> Any:
```
linjiX, line 454: shall we support int key?

graviti/utility/user.py
```diff
@@ -355,3 +357,145 @@ def update(self, __m: Any = (), **kwargs: _V) -> None:  # pylint: disable=argume
...
+        Pairs are returned in LIFO (last-in, first-out) order.
+
+        Returns:
+            A (key, value) pair as a tuple.
+
+        """
+        key = self._keys.pop()
+        return (key, self._data.pop(key))
+
+    def setdefault(self, key: str, default: _V = None) -> _V:  # type: ignore[assignment]
```
linjiX, line 485: int key?

linjiX (changes requested):

graviti/utility/user.py
Lee-000: If `items()` returns `Iterator`, should `keys()` and `values()` also return `Iterator`?
In the current situation, there wouldn't be too many items in it.

graviti/utility/user.py
Lee-000: I think it is wired to support int key here. `__setitem__ should be used to set the value at a specific index.

graviti/utility/user.py
linjiX: obviously Iterator is better than tuple

graviti/utility/user.py
```diff
@@ -355,3 +357,178 @@ def update(self, __m: Any = (), **kwargs: _V) -> None:  # pylint: disable=argume
 
         """
         self._data.update(__m, **kwargs)
+
+
+class UserOrderedDict(MutableMapping[str, _V]):
```
linjiX, line 362: can you get a better name?

graviti/utility/user.py
```diff
@@ -355,3 +357,178 @@ def update(self, __m: Any = (), **kwargs: _V) -> None:  # pylint: disable=argume
...
+
+    @overload
+    def get(self, key: str) -> Optional[_V]:  # pylint: disable=arguments-differ
+        ...
+
+    @overload
+    def get(self, key: str, default: Union[_V, _T] = ...) -> Union[_V, _T]:
+        ...
+
+    def get(self, key: str, default: Any = None) -> Any:
```
linjiX, line 464: support int key?

graviti/utility/user.py
```diff
@@ -355,3 +357,178 @@ def update(self, __m: Any = (), **kwargs: _V) -> None:  # pylint: disable=argume
...
+        if not self._data.__contains__(key):
+            self._keys.append(key)
+
+        self._data.__setitem__(key, value)
+
+    def __delitem__(self, key: Union[int, str]) -> None:
+        if isinstance(key, int):
+            key = self._keys.pop(key)
+        else:
+            self._keys.remove(key)
```
linjiX, line 408: what if the str key is not in `keys`?
what exception should we raise?

graviti/utility/user.py
linjiX: why it is wired?

linjiX (changes requested):

graviti/schema/builtin.py
```diff
@@ -261,29 +261,14 @@ class record(PortexBuiltinType):  # pylint: disable=invalid-name
           },
         )
 
-        Create a record by :class:`Field` list:
-
-        >>> t = record([Field("f0", float64(0)), Field("f1", array(int64()))])
```
linjiX, line 266: Field is also tuple, it is legal input of OrderDict, shall we put the Field remove job into another PR?

graviti/utility/user.py
```diff
@@ -355,3 +357,178 @@ def update(self, __m: Any = (), **kwargs: _V) -> None:  # pylint: disable=argume
...
+        Arguments:
+            key: The key for dict.
+            default: The value to be returned if the key is not in the dict and it is given.
+
+        Returns:
+            Value to be removed from the UserOrderedDict object.
+
+        """
+        if isinstance(key, int):
+            key = self._keys.pop(key)
```
linjiX, line 497: what if the `self._keys.pop` raise `IndexError`?
if the default is given, `pop` should not raise when the key does not exist.

linjiX (changes requested):

graviti/utility/user.py
```diff
@@ -355,3 +357,178 @@ def update(self, __m: Any = (), **kwargs: _V) -> None:  # pylint: disable=argume
...
+        elif self._data.__contains__(key):
+            self._keys.remove(key)
+
+        if default is self.__marker:
+            return self._data.pop(key)
+
+        return self._data.pop(key, default)
+
+    def popitem(self) -> Tuple[str, _V]:
```
linjiX, line 506: what exception should `popitem` raise when the dict is empty?

linjiX (changes requested):

graviti/utility/user.py
Lee-000: We only support int in those methods that exit in `list` as well currently.

graviti/utility/user.py
```diff
@@ -355,3 +357,86 @@ def update(self, __m: Any = (), **kwargs: _V) -> None:  # pylint: disable=argume
...
+    def __iter__(self) -> Iterator[str]:
+        return self._keys.__iter__()
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, self.__class__):
+            return False
+
+        return self._keys.__eq__(other._keys) and self._data.__eq__(other._data)
+
+    def popitem(self) -> Tuple[str, _V]:
```
linjiX, line 425: why we need `popitem` here?
basicly `popitem` is a remove operation, but this `popitem` has not called `__delitem__`?

linjiX (changes requested):

graviti/utility/user.py
```diff
@@ -355,3 +357,86 @@ def update(self, __m: Any = (), **kwargs: _V) -> None:  # pylint: disable=argume
 
         """
         self._data.update(__m, **kwargs)
+
+
+class StringKeyOrderedDict(MutableMapping[str, _V]):
```
linjiX, line 362: this name is too long, can we get a better name?

linjiX (changes requested):

graviti/utility/user.py
linjiX: also shall we still put this class in `user.py`?

graviti/utility/user.py
Lee-000: Because if use the `MutableMapping.popitem` method, the first item will be popped. And we want it to popitem from the tail.

linjiX: Ok, we need to call `self.__delitem__` in popitem

linjiX (changes requested):

graviti/utility/user.py
```diff
@@ -355,3 +357,88 @@ def update(self, __m: Any = (), **kwargs: _V) -> None:  # pylint: disable=argume
 
         """
         self._data.update(__m, **kwargs)
+
+
+class SimpleOrderedDict(MutableMapping[str, _V]):
```
linjiX, line 362: This name is too simple, can you get a better name?

linjiX (changes requested):

graviti/utility/user.py
```diff
@@ -355,3 +357,88 @@ def update(self, __m: Any = (), **kwargs: _V) -> None:  # pylint: disable=argume
...
+        Returns:
+            A (key, value) pair as a tuple.
+
+        """
+        try:
+            key = self._keys[-1]
+        except IndexError:
+            raise KeyError("popitem(): dictionary is empty") from None
+
+        value = self.__getitem__(key)
```
linjiX, line 442: can we use `self.pop`, `pop` calls `__delitem__`

linjiX (changes requested):

graviti/utility/user.py
```diff
@@ -355,3 +357,87 @@ def update(self, __m: Any = (), **kwargs: _V) -> None:  # pylint: disable=argume
...
+        Returns:
+            A (key, value) pair as a tuple.
+
+        """
+        try:
+            key = self._keys[-1]
+        except IndexError:
+            raise KeyError("popitem(): dictionary is empty") from None
+
+        value = self.pop(key)
```
linjiX, line 442: `return key, self.pop(key)`

linjiX (changes requested):

linjiX (approved):

PR #112, opened 2022-04-20
refactor: migrate "utility.requests" and rename "openapi.request"

wangyuqing0424:


GravitiRob:
Task linked: [CU-27ygg0f TensorBay 代码迁移：request.py 路径调整与重命名](https://app.clickup.com/t/27ygg0f)

graczhual (approved):

graviti/client/branch.py
```diff
@@ -8,7 +8,7 @@
 from typing import Any, Dict, Optional
 from urllib.parse import urljoin
 
-from graviti.utility import URL_PATH_PREFIX, open_api_do
+from graviti.client.requests import URL_PATH_PREFIX, open_api_do
```
linjiX, line 11: we have more than one commit scope in this commit: client + openapi + utility,
in this case, we can remove the scope in commit message

linjiX (changes requested):

graviti/client/branch.py
```diff
@@ -8,7 +8,7 @@
 from typing import Any, Dict, Optional
```
graczhual, line 8: commit message not clear, can we use (clientopenapi) as scope?

graviti/client/branch.py
linjiX: we have more than one requests.py file in our repo, it will be better to point out the requests is `utility/requests.py`
or `utility.requests`

linjiX (changes requested):

Lee-000 (commented):
If the pr includes modification in different modules, the scope can be omitted.

linjiX (approved):

PR #115, opened 2022-04-20
feat(schema): implement "Fields.rename" method

Lee-000:


graviti/schema/field.py
```diff
@@ -101,6 +101,17 @@ def insert(self, index: int, name: str, portex_type: PortexType) -> None:
         self._keys.insert(index, name)
         self._data.__setitem__(name, portex_type)
 
+    def rename(self, current_name: str, new_name: str) -> None:
```
linjiX, line 104: I think `new` should be a pair of `old`, shall we use `old_name` rather than `current_name`?

linjiX (approved):

PR #118, opened 2022-04-20
refactor: adapt "DatasetManager" with OpenAPIv2

graczhual:


GravitiRob:
Task linked: [CU-2k67b5k [parent] 对接基础 OpenAPIv2 与 Manager](https://app.clickup.com/t/2k67b5k)

graviti/graviti.py
```diff
@@ -6,6 +6,7 @@
 """The implementation of the Graviti."""
 
 from graviti.manager import DatasetManager
+from graviti.openapi import get_current_user
```
linjiX, line 9: The commit message is not clear

graviti/manager/dataset.py
```diff
@@ -5,34 +5,30 @@
...
 from graviti.schema import Extractors, catalog_to_schema, get_extractors
-from graviti.utility import LazyFactory, LazyList, NestedDict
+from graviti.utility import LazyFactory, LazyList, NestedDict, ReprMixin
 
 LazyLists = NestedDict[str, LazyList[Any]]
 
 
-class Dataset(Mapping[str, DataFrame]):  # pylint: disable=too-many-instance-attributes
+class Dataset(  # pylint: disable=too-many-instance-attributes
+    AttrsMixin, ReprMixin, Mapping[str, DataFrame]
```
linjiX, line 30: shall we set `_repr_type`?

graviti/manager/dataset.py
```diff
@@ -41,12 +37,44 @@ class Dataset(Mapping[str, DataFrame]):  # pylint: disable=too-many-instance-att
...
+    _name: str = attr(key="name")
+    _alias: str = attr(key="alias")
+    _default_branch: str = attr(key="default_branch")
+    _commit_id: str = attr(key="commit_id")
+    created_at: str = attr()
+    updated_at: str = attr()
+    _owner: str = attr(key="owner")
+    _is_public: bool = attr(key="is_public")
+    _config: str = attr(key="config")
+    _branch: Optional[str] = attr(key="branch")
```
linjiX, line 77: why branch is optional？

graviti/manager/dataset.py
```diff
@@ -315,47 +412,56 @@ class DatasetManager:
...
+            The created :class:`~graviti.manager.dataset.Dataset` instance.
 
         """
+        response = create_dataset(
+            self._access_key,
+            self._url,
+            name,
+            alias=alias,
+            is_public=is_public,
+            config=config,
```
linjiX, line 454: shall we put name, alias, is_public ... into a dict?

graviti/manager/dataset.py
```diff
@@ -56,18 +84,28 @@ def __init__(
...
         alias: str,
+        default_branch: str,
         commit_id: str,
-        branch_name: Optional[str] = None,
+        created_at: str,
+        updated_at: str,
+        owner: str,
+        is_public: bool,
+        config: str,
+        branch: Optional[str] = None,
```
linjiX, line 94: can we remove `branch` here just use `default_branch`?

linjiX (changes requested):

graviti/manager/dataset.py
```diff
@@ -41,12 +37,44 @@ class Dataset(Mapping[str, DataFrame]):  # pylint: disable=too-many-instance-att
...
+    )
+
+    _access_key: str = attr(key="access_key")
+    _url: str = attr(key="url")
+    _dataset_id: str = attr(key="id")
+    _name: str = attr(key="name")
+    _alias: str = attr(key="alias")
+    _default_branch: str = attr(key="default_branch")
+    _commit_id: str = attr(key="commit_id")
+    created_at: str = attr()
```
linjiX, line 72: why `created_at` is not property?

graviti/manager/dataset.py
graczhual: when we checkout to a commit, the branch will be None

graviti/manager/dataset.py
graczhual: branch is useful when we checkout to other branch

graviti/manager/dataset.py
graczhual: I'm not sure. I just don't think this attribute is important, and it doesn't matter if users change it. What should we use to determine when to use property? the attribute is not modifiable?

graviti/manager/dataset.py
linjiX: got it

graviti/manager/dataset.py
linjiX: can we just set `branch = default_branch` in `__init__` and remove the `branch` as input?

graviti/manager/dataset.py
```diff
@@ -41,12 +37,44 @@ class Dataset(Mapping[str, DataFrame]):  # pylint: disable=too-many-instance-att
...
+        updated_at: The time of last update.
+        owner: The owner of the dataset.
+        is_public: Whether the dataset is public.
+        config: The config of dataset.
+        branch: The current branch name of the dataset.
 
     """
 
     _data: Dict[str, DataFrame]
+    _T = TypeVar("_T", bound="Dataset")
```
Lee-000, line 52: Should the order of the variables follow the standard?

graviti/manager/dataset.py
graczhual: _repr_type = ReprType.INSTANCE
do not need to set here, this is the default value

linjiX: yes the key is sheet name, will it make something wrong ?

graczhual: So when we get a dataset, all segments will be initialized?

linjiX: ReprMixin only effects `__repr__`

graviti/manager/dataset.py
```diff
@@ -315,47 +432,48 @@ class DatasetManager:
...
-            yield Dataset(
-                self._access_key,
-                self._url,
-                item["id"],
-                item["name"],
-                branch_name=item["defaultBranch"],
-                commit_id=item["commitId"],
-                alias=item["alias"],
-            )
+            yield Dataset.from_pyobj(item)
```
Lee-000, line 444: Does `item` contain `accesskey` and `url`?

graviti/manager/dataset.py
```diff
@@ -315,47 +432,48 @@ class DatasetManager:
...
+        arguments: Dict[str, Any] = {
+            "name": name,
+            "alias": alias,
+            "is_public": is_public,
+            "config": config,
+        }
+        response = create_dataset(self._access_key, self._url, **arguments)
+        response.update(arguments)
+
+        return Dataset.from_pyobj(response)
```
Lee-000, line 476: Why is it necessary to add the method `from_pyobj`? Would `init` do?

graviti/manager/dataset.py
```diff
@@ -41,11 +37,44 @@ class Dataset(Mapping[str, DataFrame]):  # pylint: disable=too-many-instance-att
...
+        "created_at",
+        "updated_at",
+        "is_public",
+        "config",
+        "branch",
+    )
+
+    access_key: str = attr()
+    url: str = attr()
+    dataset_id: str = attr(key="id")
```
linjiX, line 67: we can assume that the dataset id is useless for user, shall we make it private and remove it from the _repr_attrs

graviti/graviti.py
linjiX: How about 'adapt "DatasetManager" with OpenAPIv2'

graviti/manager/dataset.py
graczhual: some response params are different from init

graviti/manager/dataset.py
graczhual: ok

linjiX (approved):

Lee-000 (approved):

PR #124, opened 2022-04-21
feat(schema): add check for the type of key and value in "Fields"

Lee-000:


GravitiRob:
Task linked: [CU-28fm45b 添加对 Fields 添加/修改的 key 与 value 类型的检查](https://app.clickup.com/t/28fm45b)

graviti/utility/user.py
```diff
@@ -378,6 +378,8 @@ def __init__(
 
         iterable = items.items() if isinstance(items, Mapping) else items
         for key, value in iterable:
+            self._check_key(key)
+            self._check_value(value)
```
linjiX, line 382: Do we need to check the value type in base class?

Lee-000: In the base class, it does not actually check the value type, because the `_check_value` method is passed. It is for the choice of the subclass to check the value or not.

graviti/utility/user.py
```diff
@@ -422,6 +428,15 @@ def __eq__(self, other: Any) -> bool:
 
         return self._keys.__eq__(other._keys) and self._data.__eq__(other._data)
 
+    @staticmethod
+    def _check_key(key: str) -> None:
+        if not isinstance(key, str):
+            raise ValueError("The key should be a string")
```
linjiX, line 434: shall we mention the class name in the error message?

graviti/schema/field.py
```diff
@@ -60,7 +60,15 @@ class Fields(NameOrderedDict[PortexType]):
...
-        super().__init__(fields)
+        super().__init__()
+
+        iterable = fields.items() if isinstance(fields, Mapping) else fields
+        for key, value in iterable:
+            self._check_key(key)
+            self._check_value(value)
+
+            self._keys.append(key)
+            self._data[key] = value
```
linjiX, line 71: can we call `self.__setitem__` in `self.__init__`?

linjiX (changes requested):

graviti/utility/user.py
```diff
@@ -393,6 +394,8 @@ def __getitem__(self, key: Union[int, str]) -> _V:
     def __setitem__(self, key: Union[int, str], value: _V) -> None:
```
linjiX, line 394: can we add a `setitem` method only accept str key.
then we can use this `setitem` in `__init__` and `__setitem__`?

graviti/schema/field.py
```diff
@@ -60,7 +60,11 @@ class Fields(NameOrderedDict[PortexType]):
     """Represents a Portex ``record`` fields dict."""
 
     def __init__(self, fields: Union[Iterable[Tuple[str, PortexType]], Mapping[str, PortexType]]):
-        super().__init__(fields)
+        super().__init__()
```
linjiX, line 63: do we still need to rewrite `__init__`?

linjiX (changes requested):

graviti/utility/user.py
```diff
@@ -422,6 +418,18 @@ def __eq__(self, other: Any) -> bool:
 
         return self._keys.__eq__(other._keys) and self._data.__eq__(other._data)
 
+    @classmethod
+    def _check_key(cls, key: str) -> None:
+        if not isinstance(key, str):
+            raise ValueError(f'The key in "{cls.__name__}" should be a string')
```
linjiX, line 424: Should it be `TypeError` or `KeyError`?

linjiX (changes requested):

linjiX (approved):

PR #126, opened 2022-04-21
refactor(schema): remove "field.Field"

Lee-000:


GravitiRob:
Task linked: [CU-2840ypd 删除 Field 类型](https://app.clickup.com/t/2840ypd)

graviti/schema/factory.py
```diff
@@ -73,13 +86,10 @@ def __call__(self, **kwargs: Any) -> PTYPE.PType:
 
         """
         ptype = self._ptype_getter(**kwargs)
-        if ptype in {PTYPE.PortexType, PTYPE.Field}:
+        if ptype in {PTYPE.PortexType}:
```
linjiX, line 89: why still using `in`?

graviti/schema/factory.py
```diff
@@ -111,7 +121,7 @@ def __call__(self, **kwargs: Any) -> PTYPE.PType:
 
         """
         if self._ptype_getter(**kwargs) == PTYPE.Fields:
-            return PTYPE.Field
+            return PTYPE.Fields
```
linjiX, line 124: The item of `Fields` should not be `Fields`

graviti/schema/factory.py
```diff
@@ -338,6 +348,9 @@ def __init__(self, decl: List[Any], ptype: PTYPE.PType = PTYPE.Any) -> None:
         dependences = set()
         keys = {}
         ptype = DynamicListParameter(ptype) if isinstance(ptype, Dynamic) else ptype
+        if ptype == PTYPE.Fields:
+            decl = [decl]
```
linjiX, line 352: why we need to do that?

linjiX (changes requested):

graviti/schema/factory.py
```diff
@@ -398,53 +411,6 @@ def __call__(self, **kwargs: Any) -> Dict[str, Any]:
         return {key: factory(**kwargs) for key, factory in self._factories.items()}
 
 
-class FieldFactory(Factory):
```
linjiX, line 401: Why not keep `FieldFactory`?

linjiX (changes requested):

graviti/schema/ptype.py
```diff
@@ -132,44 +131,54 @@ class Field(ParameterType):
...
-            A :class:`Field` instance created by the input argument.
+            A tuple of str and PortexType created by the input argument.
 
         """
-        return ClassField(*arg)
+        try:
+            name, portex_type = arg
+            if isinstance(name, str) and isinstance(portex_type, ClassPortexType):
+                return name, portex_type
+        except ValueError:
```
linjiX, line 151: why we pass the ValueError?

Lee-000: TypeError will be raised instead.

linjiX: Is that necessary?

linjiX (approved):

PR #127, opened 2022-04-21
refactor(portex): load Param default value according to the ptype

Lee-000:


GravitiRob:
Task linked: [CU-2845rtv 添加默认值 Fields() 至未传入的类型为 Fields 的参数](https://app.clickup.com/t/2845rtv)

graviti/schema/base.py
```diff
@@ -47,37 +48,44 @@ def _repr1(self, level: int) -> str:
...
+                f"{attr._repr1(level + 1) if hasattr(attr, '_repr1') else repr(attr)},"
+            )
 
         if with_params:
             lines.append(")")
             return f"\n{indent}".join(lines)
 
         return f"{lines[0]})"
 
+    def _generate_arguments_and_parameters(
```
linjiX, line 65: this method is too complex

linjiX (changes requested):

graviti/schema/base.py
```diff
@@ -47,37 +48,42 @@ def _repr1(self, level: int) -> str:
         with_params = False
         indent = level * _INDENT
         lines = [f"{self.__class__.__name__}("]
-        for name, parameter in self.params.items():
-            attr = getattr(self, name)
-            if attr != parameter.default:
```
linjiX, line 52: If we add a method `Param.check_default` to check wether the attr is equal to its default value, and handle the Fields logic in it?
Will it help?

graviti/schema/param.py
```diff
@@ -102,6 +103,21 @@ def to_pyobj(self) -> Dict[str, Any]:
 
         return pyobj
 
+    def is_default(self, arg: Any) -> Any:
```
linjiX, line 106: why the return type is Any？

graviti/schema/param.py
```diff
@@ -116,6 +132,9 @@ def check(self, arg: Any) -> Any:
             ValueError: Raise when the argument is not in options.
 
         """
+        if not arg and self.ptype == PTYPE.Fields:
+            return Fields()
+
         if arg == self.default:
```
linjiX, line 138: shall we call `is_default` here?

linjiX: can we check like this?
```python
if self.ptype != PTYPE.Fields and arg == self.default:
```
so that the None value will pass the default check and be sent into `ptype.check`

Lee-000: We need to know the `ptype` to return the value. And no need to check the default value. So `is_default` is not called here.

graviti/portex/builder.py
```diff
@@ -251,11 +251,10 @@ def build(self) -> Type["PortexExternalType"]:
         factory = TypeFactory(decl, imports)
 
         keys = factory.keys
+        for param_pyobj in params_pyobj:
+            param_pyobj["ptype"] = keys.get(param_pyobj["name"], PTYPE.Any)
```
linjiX, line 255: will it be better to make key as a parameter of `Params.from_pyobj`?

graviti/portex/ptype.py
```diff
@@ -286,7 +294,9 @@ def load(content: Dict[str, TypingAny], imports: Optional[Imports] = None) -> Cl
             A Portex type instance created from the input python dict.
 
         """
-        return ClassPortexType.from_pyobj(content, imports)
+        if content is not None:
+            return ClassPortexType.from_pyobj(content, imports)
+        return content
```
linjiX, line 299: under what condition the portex type will be null?

Lee-000: `visible` in keypoints.

linjiX (approved):

PR #132, opened 2022-04-24
feat: adapt "CommitManager" with OpenAPIv2

graczhual:


GravitiRob:
Task linked: [CU-2k67b79 Manager 对接：Commit](https://app.clickup.com/t/2k67b79)

graviti/manager/commit.py
```diff
@@ -191,58 +189,55 @@ class CommitManager:
...
 
         """
-        try:
-            commit = next(self._generate(revision))
-        except StopIteration as error:
-            raise ResourceNotExistError(resource="commit", identification=revision) from error
+        if revision is None:
+            commit_id = self._commit_id
+        else:
+            commit_id = get_revision(**self._access_info, revision=revision)["commit_id"]
```
linjiX, line 235: Should `get_revision` returns all commit info? if yes, why we need to call `get_commit`?

linjiX (changes requested):

graviti/manager/commit.py
graczhual: `get_revision` returns all commit info. we do not need to call get_commit here.  So we do not need api `get commit`

linjiX (approved):

PR #133, opened 2022-04-24
refactor(schema): return the builtin PyArrow type in "to_pyarrow"

Lee-000:


GravitiRob:
Task linked: [CU-28fkztb 增加 Portex schema 转 PyArrow schema 的完全展开形式](https://app.clickup.com/t/28fkztb)

graviti/schema/base.py
```diff
@@ -172,14 +172,14 @@ def to_yaml(self) -> str:
         """
         return yaml.dump(self.to_pyobj(), sort_keys=False)  # type: ignore[no-any-return]
 
-    def to_pyarrow(self) -> ExtensionBase:
-        """Convert the instance to an ``ExtensionBase`` instance.
+    def get_pyarrow_type(self) -> pa.DataType:
```
linjiX, line 175: can we use "to_pyarrow" as the method name?

graviti/schema/field.py
```diff
@@ -125,11 +125,11 @@ def to_pyobj(self) -> List[Dict[str, Any]]:
         """
         return [{"name": name, **portex_type.to_pyobj(False)} for name, portex_type in self.items()]
 
-    def to_pyarrow(self) -> List[pa.Field]:
-        """Convert the fields to a list of PyArrow fields.
+    def to_pyarrow(self) -> List[pa.DataType]:
```
linjiX, line 128: should the return type be `List[pa.Field]`?

graviti/schema/builtin.py
```diff
@@ -32,7 +28,7 @@
 }
 
 
-class PortexBuiltinType(PortexType):
+class PortexBuiltinType(PortexType):  # pylint: disable=abstract-method
```
linjiX, line 31: where the pylint warning comes from?

graviti/schema/builtin.py
```diff
@@ -273,7 +297,13 @@ def __init__(
...
 
-    def _get_pyarrow_type(self) -> "paDataType":
+    def to_pyarrow(self) -> pa.DataType:
+        """Convert the Portex type to the corresponding builtin PyArrow DataType.
+
+        Returns:
+            The corresponding builtin PyArrow DataType.
+
+        """
         return pa.struct(self.fields.to_pyarrow())  # pylint: disable=no-member
```
linjiX, line 307: where the pylint warning comes from?

linjiX (changes requested):

graviti/schema/builtin.py
Lee-000: Because this class does not implement the `to_pyarrow` method which is not implemented in its parent class.

graviti/schema/builtin.py
Lee-000: Pylint considers `self.fields` as a tuple, which does not have `to_pyarrow` method.

linjiX (approved):

PR #135, opened 2022-04-24
feat: adapt "TagManager" with OpenAPIv2

graczhual:


GravitiRob:
Task linked: [CU-2k67b8d Manager 对接：Tag](https://app.clickup.com/t/2k67b8d)

graviti/manager/tag.py
```diff
@@ -51,34 +64,55 @@ def create(self, name: str, revision: Optional[str] = None) -> TensorbayTag:
...
         """Delete a tag.
 
         Arguments:
             name: The tag name to be deleted for the specific commit.
 
+        Raises:
+            TypeError: When the given tag name is illegal.
+
         """
+        if not name:
```
linjiX, line 115: why need to check name here?
can OpenAPI check this for us?

graczhual: because here the name is the part of url. If the name = "", the url will be change, then backend will raise ResourceNotExist error about the url instead of the ResourceNotExist about tag

linjiX: OK,
1. Should we move the check to `openapi` level?
2. Should we also check the name is `str` or not?

graczhual: 1. yes
2. No, in the openapi, we use f string. Here we remind the user of the variable type, and even if the user set name = 1. The openapi will not raise error. I think this way is ok.

graviti/manager/tag.py
```diff
@@ -51,34 +64,55 @@ def create(self, name: str, revision: Optional[str] = None) -> TensorbayTag:
...
-            The :class:`.TensorbayTag` instance with the given name.
+        Raises:
+            TypeError: When the given tag name is illegal.
+
+        Returns:
+            The :class:`.Tag` instance with the given name.
 
         """
+        if not name:
+            raise TypeError("The given tag name is illegal")
```
linjiX, line 91: should it raise `TypeError`?

linjiX (approved):

PR #141, opened 2022-04-25
feat(portex): add "record.get_keys" method

Lee-000:


GravitiRob:
Task linked: [CU-2d3adbu 实现 "PortexType.get_keys" 方法](https://app.clickup.com/t/2d3adbu)

graviti/schema/builtin.py
```diff
@@ -306,6 +307,26 @@ def to_pyarrow(self) -> pa.DataType:
...
+
+        Returns:
+            The keys to create lazy lists.
+
+        """
+        keys = []
+        for name, portex_type in self.fields.items():  # pylint: disable=no-member
+            key = (name,)
+            if isinstance(portex_type, PortexExternalType):
+                portex_type = portex_type.internal_type
```
linjiX, line 321: what if the `internal_type` is another `PortexExternalType`?

graviti/schema/builtin.py
```diff
@@ -306,6 +307,26 @@ def to_pyarrow(self) -> pa.DataType:
...
+        """Get the keys of each field to create lazy lists.
+
+        Returns:
+            The keys to create lazy lists.
+
+        """
+        keys = []
+        for name, portex_type in self.fields.items():  # pylint: disable=no-member
+            key = (name,)
+            if isinstance(portex_type, PortexExternalType):
```
linjiX, line 320: why not implement get_keys for `PortexExternalType`? even all `PortexType`?

linjiX (changes requested):

graviti/schema/base.py
```diff
@@ -183,3 +183,14 @@ def to_pyarrow(self) -> pa.DataType:
 
         """
         raise NotImplementedError
+
+    def get_keys(  # pylint: disable=no-self-use
+        self,
+    ) -> Union[Tuple[str, ...], List[Tuple[str, ...]]]:
```
linjiX, line 189: why the return type is a union of `List` and `Tuple`, can we return Tuple for all get_keys method?

linjiX (changes requested):

graviti/schema/builtin.py
```diff
@@ -306,6 +306,25 @@ def to_pyarrow(self) -> pa.DataType:
         """
         return pa.struct(self.fields.to_pyarrow())  # pylint: disable=no-member
 
+    def get_keys(self) -> List[Tuple[str, ...]]:
+        """Get the keys of each field to create lazy lists.
+
+        Returns:
+            The keys to create lazy lists.
```
linjiX, line 313: paging list?
Also I think we can use the keys to do other things not only create paging list.
Should we mention paging list in the doc?

linjiX (changes requested):

graviti/schema/base.py
Lee-000: The return type of `record.get_keys()` is `List`, since there are multiple names in the record.

graviti/portex/builtin.py
```diff
@@ -306,6 +306,28 @@ def to_pyarrow(self) -> pa.DataType:
...
+            sub_keys = portex_type._get_keys()  # pylint: disable=protected-access
+            if sub_keys:
+                for sub_key in sub_keys:
+                    keys.append(key + sub_key)
+            else:
+                keys.append(key)
+
+        return keys
+
+    def get_keys(self) -> List[Tuple[Any, ...]]:
```
linjiX, line 322: Why `Any`? should the keys be `str`?

linjiX (changes requested):

graviti/portex/builtin.py
Lee-000: Because the keys could be nested.

graviti/portex/builtin.py
```diff
@@ -306,6 +306,28 @@ def to_pyarrow(self) -> pa.DataType:
         """
         return pa.struct(self.fields.to_pyarrow())  # pylint: disable=no-member
 
+    def _get_keys(self) -> List[Tuple[Any, ...]]:
+        keys = []
+        for name, portex_type in self.fields.items():  # pylint: disable=no-member
+            key = (name,)
+            sub_keys = portex_type._get_keys()  # pylint: disable=protected-access
```
linjiX, line 313: seems `subkey` is a real word.
should we remove the `_`?

linjiX (changes requested):

linjiX (approved):

PR #155, opened 2022-04-27
feat(manager): implement the "commit" method in class "Draft"

graczhual:


GravitiRob:
Task linked: [CU-290mfy4 实现 Draft.commit 方法](https://app.clickup.com/t/290mfy4)

graviti/manager/draft.py
```diff
@@ -93,14 +94,29 @@ def close(self) -> None:
         self.state = "CLOSED"
         # TODO: update the draft.updated_at
 
-    def commit(self, title: str, description: str = "") -> None:
+    def commit(self, title: str, description: Optional[str] = None) -> Commit:
```
linjiX, line 97: why change the default description to `None`?

linjiX (changes requested):

graviti/manager/draft.py
graczhual: OpenAPI has a default value for description. Here, we do not need to set default value equal to "". If user do not write description, we don't need to pass this parameter.

linjiX (approved):

PR #159, opened 2022-04-28
feat(portex): add "PortexExternalType.to_builtin" method

Lee-000:


GravitiRob:
Task linked: [CU-29htfaq 实现 PortexType.to_builtin 方法](https://app.clickup.com/t/29htfaq)

graviti/portex/template.py
```diff
@@ -60,6 +61,19 @@ def to_pyarrow(self) -> pa.DataType:
         """
         return self.internal_type.to_pyarrow()
 
+    def to_builtin(self) -> PortexBuiltinType:
+        """Expand the top level of the Portex external type to Portex builtin type.
+
+        Returns:
+            The expanded Portex builtin type.
+
+        """
```
linjiX, line 70: Do we need recursion here?
```python
type = self
while not isinstance(type, PortexBuiltinType):
    type = type.internal_type

return type
```

linjiX (changes requested):

graviti/portex/builtin.py
```diff
@@ -56,6 +56,15 @@ def __init__(self, **kwargs: Any) -> None:
     def _from_pyarrow(cls: Type[_T], pyarrow_type: pa.DataType) -> _T:
         return cls()
 
+    def to_builtin(self: _T) -> _T:
```
linjiX, line 59: do we need to define `to_builtin` method in `PortexType`?

Lee-000: The return type of `to_builtin` is `PortexBuiltinType`. I don't think it is appropriate to define this in the parent class.

linjiX (approved):

PR #161, opened 2022-04-28
feat(portex): implement "PortexType.copy"

linjiX:


GravitiRob:
Task linked: [CU-29hxbnt 实现 PortexType.copy 方法](https://app.clickup.com/t/29hxbnt)

graviti/portex/field.py
```diff
@@ -5,14 +5,16 @@
...
-from typing import Any, Dict, Iterable, List, Mapping, Tuple, Union
+from typing import Any, Dict, Iterable, List, Mapping, Tuple, TypeVar, Union
 
 import pyarrow as pa
 
 from graviti.portex.base import _INDENT, PortexType
 from graviti.portex.package import Imports
 from graviti.utility import NameOrderedDict
 
+_T = TypeVar("_T", bound="Fields")
```
Lee-000, line 16: Where should the `TypeVar` be? Inside or outside of the class? Any differences?

linjiX: If the TypeVar is inside the class, it cannot be used in the class method.

linjiX: And inside the class will give the TypeVar a more explicit scope.

Lee-000 (approved):

PR #169, opened 2022-04-29
feat(dataframe): stop supporting "column.Series.__init__"

AChenQ:


GravitiRob:
Task linked: [CU-28v2d3k 实现 DataFrame.extend 方法](https://app.clickup.com/t/28v2d3k)

graviti/dataframe/column/series.py
```diff
@@ -43,20 +43,16 @@ class Series:
     """
 
     schema: PortexType
+    _data: PagingList
+    name: Optional[str]
```
linjiX, line 47: the commit type should be `feat`, because stop some feature also effect the features.

graviti/dataframe/column/series.py
```diff
@@ -43,20 +43,16 @@ class Series:
...
 
     def __init__(
         self,
         data: Sequence[Any],
         schema: Any = None,
         name: Union[str, int, None] = None,
```
linjiX, line 53: should we update the signature?

linjiX (changes requested):

PR #175, opened 2022-05-02
docs(openapi): modify the output "schema" of Examples in "sheet.py"

graczhual:


graviti/openapi/data.py
```diff
@@ -55,9 +55,9 @@ def list_draft_data(
         sheet: The sheet name.
         columns: The string of column names separated by ``|``.
             Multiple indexes can be expressed using dots. None means to get all columns.
-        order_by: The string determine the order of sorting by the order in which column
-            names appear. If None or incomplete, the remainder is sorted by the primary key
-            of the sheet.
+        order_by: The string of column names whose order determines the precedence of the sort.
```
linjiX, line 58: the commit message is not clear

linjiX (changes requested):

graviti/openapi/data.py
```diff
@@ -55,9 +55,9 @@ def list_draft_data(
...
+            If the value is None or incomplete, the remainder will be sorted by the primary
```
Lee-000, line 59: What does `incomplete` means? What kind of argument is `incomplete`?

graczhual: For example, columns = "filename, category". Here order_by = "filename"

linjiX (approved):

PR #176, opened 2022-05-02
feat(utility): rename File to RemoteFile and add File for local files

Lee-000:


GravitiRob:
Task linked: [CU-29xepjd 实现 File 类用于代表本地文件](https://app.clickup.com/t/29xepjd)

graviti/utility/file.py
```diff
@@ -54,3 +56,61 @@ def checksum(self) -> str:
 
         """
         return self._checksum
+
+
+class LocalFile(ReprMixin):
```
linjiX, line 61: should `LocalFile` and `RemoteFile` share the same base class？

linjiX: And this class will be directly used by customer? Is `LocalFile` a good name?

linjiX (approved):

PR #191, opened 2022-05-03
feat: adapt to the change in the response of OpenAPIv2 "Create dataset"

graczhual:


graviti/manager/dataset.py
```diff
@@ -240,15 +240,10 @@ def create(
...
-        arguments: Dict[str, Any] = {
-            "access_key": self.access_key,
-            "url": self.url,
-            "name": name,
-            "alias": alias,
-            "is_public": is_public,
-            "config": config,
-        }
-        response = create_dataset(**arguments)
+        arguments = {"access_key": self.access_key, "url": self.url}
```
linjiX, line 243: "adapt to the change in xxx" is not clear

linjiX (changes requested):

linjiX (approved):

PR #196, opened 2022-05-04
fix(utility): remove "encoding" in open with "rb" mode to fix ValueError

Lee-000:


graviti/utility/file.py
```diff
@@ -148,7 +148,7 @@ def checksum(self) -> str:
         """
         if not hasattr(self, "_checksum"):
             sha1_object = sha1()
-            with self.path.open("rb", encoding="utf-8") as fp:
+            with self.path.open("rb") as fp:
```
linjiX, line 151: better to add double quote in commit message

linjiX (approved):

AChenQ (approved):

PR #210, opened 2022-05-05
fix: decode the PyArrow schema bytes as string to be JSON serializable

Lee-000:


linjiX (approved):

graviti/openapi/sheet.py
```diff
@@ -51,7 +51,7 @@ def create_sheet(
     name: str,
     schema: str,
     _avro_schema: str,
-    _arrow_schema: Optional[ByteString] = None,
+    _arrow_schema: Optional[str] = None,
     record_key_strategy: Optional[str] = None,
 ) -> None:
     """Execute the OpenAPI `POST /v2/datasets/{owner}/{dataset}/drafts/{draft_number}/sheets`.
```
graczhual, line 57: modify the Dict[str, Any] in post data

graczhual (changes requested):

graczhual (approved):

PR #213, opened 2022-05-06
fix(openapi): change the boolean value in params to JSON string

Lee-000:


graviti/openapi/requests.py
```diff
@@ -27,6 +28,10 @@ def do(method: str, url: str, **kwargs: Any) -> Response:  # pylint: disable=inv
         Response of the request.
 
     """
+    params = kwargs.get("params", {})
```
linjiX, line 31: shall we set `params` as a separate keyword-only argument?

graczhual (approved):

Lee-000:
There is no standard for URL query, so OpenAPI should adapt this behavior according to the design instead of SDK.

PR #215, opened 2022-05-06
fix(utility): correct the namespace of list items in Avro schema

Lee-000:


graviti/utility/avro.py
```diff
@@ -145,7 +145,7 @@ def to_json(self):
 
 
 def _on_list(names, namespace, name, _pa_ist: pa.ListType) -> AvroArraySchema:
-    sub_namespace = f"{namespace}.{name}"
+    sub_namespace = f"{namespace}.{name}.items"
```
linjiX, line 148: should the commit type be `fix`?

linjiX (approved):

PR #219, opened 2022-05-07
feat: add update to "updated_at" to "commit", "edit", "close" in "Draft"

graczhual:


graviti/manager/draft.py
```diff
@@ -139,24 +139,22 @@ def edit(self, title: Optional[str] = None, description: Optional[str] = None) -
...
-            self.description = description
-        # TODO: update the draft.updated_at
+        self.title = response["title"]
+        self.description = response["description"]
+        self.updated_at = response["updated_at"]
 
     def close(self) -> None:
         """Close the draft."""
-        update_draft(
+        response = update_draft(
```
linjiX, line 148: should it be `draft_info` or use `response` for all the api return

linjiX (changes requested):

graviti/manager/draft.py
graczhual: should use response for all the api return

linjiX (approved):

PR #225, opened 2022-05-09
feat(dataframe): implement "DataFrame.__setitem__"

AChenQ:


GravitiRob:
Task linked: [CU-28v2d5y 实现 DataFrame.__setitem__ 方法，支持列操作](https://app.clickup.com/t/28v2d5y)

graviti/dataframe/frame.py
```diff
@@ -126,7 +126,26 @@ def __getitem__(self, key: Union[str, Iterable[str]]) -> Container:
         raise KeyError(key)
 
     def __setitem__(self, key: str, value: Iterable[Any]) -> None:
-        pass
+        # self.schema.to_builtin always returns record.
+        field = self.schema.to_builtin().fields[key]  # type: ignore[attr-defined]
```
linjiX, line 130: will getting schema from columns be better?
```python
self[key].schema
```

graviti/dataframe/frame.py
```diff
@@ -126,7 +126,26 @@ def __getitem__(self, key: Union[str, Iterable[str]]) -> Container:
...
+            )
+        if len(column) != len(self):
+            raise ValueError(
+                f"Length of values ({len(column)}) does not match length of "
+                f"self DataFrame ({len(self)})"
+            )
+
+        self._columns[key] = column
+        self._column_names.append(key)
+        # TODO: Need add corresponding operations.
```
linjiX, line 148: also need update schema

graviti/dataframe/frame.py
linjiX: What if adding a new column rather than replace existing column?

linjiX (changes requested):

graviti/dataframe/frame.py
```diff
@@ -126,7 +126,26 @@ def __getitem__(self, key: Union[str, Iterable[str]]) -> Container:
         raise KeyError(key)
 
     def __setitem__(self, key: str, value: Iterable[Any]) -> None:
-        pass
+        # self.schema.to_builtin always returns record.
+        field = self.schema.to_builtin().fields[key]  # type: ignore[attr-defined]
+        if isinstance(value, Container):
+            if not value.schema.to_pyarrow().equals(field.to_pyarrow()):
+                raise TypeError("The schema is mismatched with the given value.")
```
linjiX, line 133: I think is acceptable to replace the column with another column with different schema

linjiX (changes requested):

graviti/dataframe/frame.py
```diff
@@ -125,8 +125,30 @@ def __getitem__(self, key: Union[str, Iterable[str]]) -> Container:
 
         raise KeyError(key)
 
-    def __setitem__(self, key: str, value: Iterable[Any]) -> None:
-        pass
+    def __setitem__(self, key: str, value: Any) -> None:
+        # self.schema.to_builtin always returns record.
+        if isinstance(value, Container):
+            field = value.schema.copy()
```
linjiX, line 131: I think `field` should be `name` + `type`. But here the `field` is just `type`.

graviti/dataframe/frame.py
```diff
@@ -125,8 +125,30 @@ def __getitem__(self, key: Union[str, Iterable[str]]) -> Container:
...
+        else:
+            if key in self._columns:
+                field = self[key].schema
+                column = field.container._from_pyarrow(self._pylist_to_pyarrow(value, field), field)
+            else:
+                array = pa.array(value)
+                field = pt.PortexType.from_pyarrow(array.type)
+                column = field.container._from_pyarrow(array, field)
+
+        if len(column) != len(self):
```
linjiX, line 142: can we check the length earlier?

linjiX (changes requested):

graviti/dataframe/frame.py
AChenQ: We cannot know the length of iterable value。

graviti/dataframe/frame.py
```diff
@@ -125,8 +126,32 @@ def __getitem__(self, key: Union[str, Iterable[str]]) -> Container:
...
         raise KeyError(key)
 
-    def __setitem__(self, key: str, value: Iterable[Any]) -> None:
-        pass
+    def __setitem__(self, key: str, value: Union[Iterable[Any], _C]) -> None:
+        # self.schema.to_builtin always returns record.
+        if isinstance(value, Container):
+            schema = value.schema.copy()
+            column = value.copy()
+        else:
```
linjiX, line 134: `elif`？

linjiX (changes requested):

linjiX (approved):

Lee-000 (approved):

PR #226, opened 2022-05-09
refactor(portex): use parameter "items" instead of "dtype" in "tensor"

Lee-000:


GravitiRob:
Task linked: [CU-2h0zf1w Portex schema 使用 items 字段更新 tensor 类型](https://app.clickup.com/t/2h0zf1w)

PR #229, opened 2022-05-11
fix(openapi): fix TypeError when comparing current time and "expire_at"

Lee-000:


graviti/openapi/file.py
```diff
@@ -57,7 +57,8 @@ def _get_upload_permission(
     sheet: str,
 ) -> Dict[str, Any]:
     key = (access_key, dataset, draft_number)
-    if int(time()) >= _PERMISSIONS.get(key, _DEFAULT_PERMISSION)["expire_at"]:
+    expire_at = _PERMISSIONS.get(key, _DEFAULT_PERMISSION)["expire_at"]
+    if expire_at == 0 or int(time()) >= convert_ISO_to_timestamp(expire_at):
```
linjiX, line 61: can we use `datetime` library for the timestamp comparison?
```python
import datetime

datetime.datetime.now()
datetime.datetime.fromisoformat()
```

linjiX (changes requested):

graczhual (approved):

graviti/portex/base.py
```diff
@@ -229,7 +229,7 @@ def to_yaml(self) -> str:
             A YAML representation of the Portex type.
 
         """
-        return yaml.dump(self.to_pyobj(), sort_keys=False)
+        return yaml.dump(self.to_pyobj(), sort_keys=False)  # type: ignore[no-any-return]
```
graczhual, line 232: why we need to add this

graviti/openapi/file.py
```diff
@@ -57,7 +57,8 @@ def _get_upload_permission(
     sheet: str,
 ) -> Dict[str, Any]:
     key = (access_key, dataset, draft_number)
-    if int(time()) >= _PERMISSIONS.get(key, _DEFAULT_PERMISSION)["expire_at"]:
+    expire_at = _PERMISSIONS.get(key, _DEFAULT_PERMISSION)["expire_at"]
+    if expire_at == 0 or datetime.now() >= datetime.fromisoformat(expire_at.rstrip("Z")):
```
linjiX, line 61: should the default `expired_at` still be `0`

linjiX: Why `rstrip("Z")`?

linjiX (changes requested):

graviti/openapi/file.py
Lee-000: Because the ISO string in response has a "Z" at the end, which is not compatible with the input format of `datetime.fromisoformat`.

linjiX (changes requested):

linjiX (approved):

PR #230, opened 2022-05-11
refactor(openapi): use aware datetime objects when comparing time

Lee-000:


graviti/openapi/file.py
```diff
@@ -43,7 +43,9 @@ def _request_upload_permission(
         is_internal=config.is_internal,
         expired=_EXPIRED_IN_SECOND,
     )
-    permission["expire_at"] = datetime.fromisoformat(permission.pop("expire_at").rstrip("Z"))
+    permission["expire_at"] = datetime.fromisoformat(
+        permission.pop("expire_at").replace("Z", "+00:00")
```
linjiX, line 47: replace look very slow, do we have faster ways?

linjiX (approved):

graviti/openapi/file.py
Lee-000: So far not found.

graczhual (approved):

PR #234, opened 2022-05-13
feat(portex): add support for object unpack grammar of Portex

Lee-000:


GravitiRob:
Task linked: [CU-2h0zej3 Portex schema 支持字典解包语法](https://app.clickup.com/t/2h0zej3)

graviti/portex/param.py
```diff
@@ -119,7 +119,7 @@ def check(self, arg: Any) -> Any:
 
         """
         if arg == self.default:
-            return arg
+            return self.ptype.load(arg)  # type: ignore[union-attr]
```
linjiX, line 122: what if the default value can not be loaded?

graviti/portex/factory.py
```diff
@@ -249,6 +267,7 @@ def __call__(self, **kwargs: Any) -> PortexType:
 
         """
         type_kwargs = {key: factory(**kwargs) for key, factory in self._factories.items()}
+        type_kwargs.update(kwargs.get(self._unpack, {}))
```
linjiX, line 270: Does that mean the unpacked param as higher priority than other param?

graviti/portex/factory.py
```diff
@@ -495,19 +540,19 @@ def __call__(self, **kwargs: Any) -> Fields:
...
     Arguments:
         decl: A dict which indicates a portex type or a dynamic portex type.
         imports: The :class:`Imports` instance to specify the import scope of the template.
 
     Returns:
-        A ``TypeFactory`` or a ``DynamicTypeFactory`` instance according to the input.
+        A ``TypeFactory`` or a ``DictUnpackTypeFactory`` instance according to the input.
 
     """
-    if decl["type"].startswith("$params."):
```
linjiX, line 509: which means the DynamicType is blocked after this commit?
should we block it after implementing all unpack grammars?

linjiX (changes requested):

graviti/portex/ptype.py
```diff
@@ -196,7 +232,9 @@ def check(arg: TypingAny) -> Tuple[str, ClassPortexType]:
         raise TypeError("Argument should be a tuple of a str and a PortexType")
 
     @staticmethod
-    def load(content: Dict[str, TypingAny], imports: Imports) -> Tuple[str, ClassPortexType]:
+    def load(
+        content: Dict[str, TypingAny], imports: Optional[Imports] = None
```
linjiX, line 236: Why we need to change the `imports` to `Optional`?

graviti/portex/factory.py
```diff
@@ -495,19 +540,19 @@ def __call__(self, **kwargs: Any) -> Fields:
...
-        return DynamicTypeFactory(decl, imports)
+    if "type" not in decl and "+" in decl:
```
linjiX, line 554: what if there are no "$params" after "+"?

linjiX (changes requested):

graviti/portex/factory.py
```diff
@@ -422,6 +477,7 @@ class FieldFactory(Factory):
     def __init__(self, decl: Dict[str, Any], imports: Imports) -> None:
         self.creator: Callable[..., Tuple[str, PortexType]]
 
+        print("FieldFactory decl", decl)
```
linjiX, line 480: remove the debug code

graviti/portex/factory.py
```diff
@@ -455,7 +511,10 @@ def __call__(self, **kwargs: Any) -> Optional[Tuple[str, PortexType]]:
         if not self._expression(**kwargs):
             return None
 
-        return self._name_factory(**kwargs), self._type_factory(**kwargs)
+        return (
+            self._name_factory(**kwargs),
+            self._type_factory(**kwargs),  # type: ignore[return-value]
```
linjiX, line 516: where is the ignore from?

graviti/portex/ptype.py
```diff
@@ -172,6 +172,42 @@ def copy(arg: _T) -> _T:
         return copy(arg)
 
 
+class ParamDict(ParameterType):
```
linjiX, line 175: can use just use `Dict` or `Mapping` rather than `ParamDict`?

linjiX (changes requested):

graviti/portex/factory.py
```diff
@@ -284,6 +302,37 @@ def __call__(self, **kwargs: Any) -> PortexType:
...
+        """Apply the input arguments to the type template.
+
+        Arguments:
+            kwargs: The input arguments.
+
+        Returns:
+            The applied Portex type.
+
+        """
+        return kwargs.get(self._key, {})
```
linjiX, line 333: why using `get` here, can we use `__getitem__` instead?

linjiX (changes requested):

graviti/portex/factory.py
```diff
@@ -284,6 +302,37 @@ def __call__(self, **kwargs: Any) -> PortexType:
         return TypeFactory(decl, self._imports)(**kwargs)
 
 
+class DictUnpackFactory(Factory):
```
linjiX, line 305: do we really need a `DictUnpackFactory`, it looks like a `VariableFactory`

linjiX (changes requested):

graviti/portex/factory.py
```diff
@@ -455,7 +485,10 @@ def __call__(self, **kwargs: Any) -> Optional[Tuple[str, PortexType]]:
         if not self._expression(**kwargs):
             return None
 
-        return self._name_factory(**kwargs), self._type_factory(**kwargs)
+        return (
+            self._name_factory(**kwargs),
+            self._type_factory(**kwargs),
+        )
```
linjiX, line 491: seems nothing changes here?

linjiX (changes requested):

graviti/portex/factory.py
```diff
@@ -493,19 +524,43 @@ def __call__(self, **kwargs: Any) -> Fields:
         )
 
 
+def object_unpack_factory_creator(decl: str, is_ptype: bool = False) -> VariableFactory:
```
linjiX, line 527: will mapping `better` than `object`

graviti/portex/factory.py
```diff
@@ -493,19 +524,43 @@ def __call__(self, **kwargs: Any) -> Fields:
...
+        is_ptype: Whether the object unpack indicates a Portex type.
+
+    Raises:
+        ValueError: When the object unpack grammar is incorrect.
+
+    Returns:
+        A ``VariableFactory`` instance or None according to the input.
+
+    """
+    if not decl.startswith("$params."):
```
linjiX, line 541: is that possible to put a real mapping after `"+"` rather than use "$params."?

linjiX (changes requested):

graviti/portex/factory.py
linjiX: Also will use `ptype` directly rather than `bool`?

linjiX (changes requested):

graviti/portex/factory.py
Lee-000: I don't see that necessary.

graviti/portex/factory.py
```diff
@@ -388,6 +409,12 @@ def __init__(self, decl: Dict[str, Any], ptype: PTYPE.PType = PTYPE.Any) -> None
         dependences = set()
         keys = {}
         for key, value in decl.items():
+            if key == "+":
```
linjiX, line 412: will it be better to check `“+”` first? rather than check it in loop?

linjiX (changes requested):

graviti/portex/factory.py
```diff
@@ -387,6 +408,13 @@ def __init__(self, decl: Dict[str, Any], ptype: PTYPE.PType = PTYPE.Any) -> None
         factories = {}
         dependences = set()
         keys = {}
+
+        unpack_value = decl.pop("+")
```
linjiX, line 412: will `pop` effect input value `decl`?

linjiX (changes requested):

linjiX (approved):

PR #235, opened 2022-05-13
fix(openapi): adapt to error response without "code" in content

Lee-000:
When requesting to openAPI, there may be exceptions during transmission,
which leads to error response without "code" in content.

### DIFF
--- a/graviti/exception.py
+++ b/graviti/exception.py
@@ -6,7 +6,7 @@
 """Basic concepts of Graviti custom exceptions."""
 
 from subprocess import CalledProcessError
-from typing import Dict, Optional, Type, TypeVar
+from typing import Dict, Optional, Tuple, Type, TypeVar
 
 from requests.models import Response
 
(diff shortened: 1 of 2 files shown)


graviti/openapi/requests.py
```diff
@@ -55,5 +55,5 @@ def open_api_do(method: str, access_key: str, url: str, **kwargs: Any) -> Respon
         return do(method=method, url=url, **kwargs)
     except ResponseError as error:
         response = error.response
-        error_code = response.json()["code"]
+        error_code = response.json().get("code", "InternalServerError")
```
linjiX, line 58: shall we use a new Error code for status code `503`
Also, shall we also handle `404` error here?

linjiX (changes requested):

graviti/exception.py
```diff
@@ -193,8 +193,13 @@ class RequestParamsMissingError(ResponseError):
     """This class defines the exception for request parameters missing response error."""
 
 
+@ResponseErrorRegister(404)
+class NotFound(ResponseError):
```
linjiX, line 197: shall we name this error as `NotFoundError`?

linjiX (approved):

graczhual (approved):

PR #236, opened 2022-05-13
fix(dataframe): fix TypeError when create a MultiIndex DataFrame

AChenQ:
The error scene:
```python
import graviti.portex as pt
from graviti import DataFrame
schema = pt.record(
    {
        "attribute": pt.record(
            {
                "truncated": pt.boolean(),
            }
        )
    }
)
data = [
    {"attribute": {"truncated": False}},
    {"attribute": {"truncated": False}}
]
df = DataFrame(data=data, schema=schema)
```
TypeError: string indices must be integers

### DIFF
--- a/graviti/dataframe/frame.py
+++ b/graviti/dataframe/frame.py
@@ -621,37 +621,30 @@ class DataFrame(Container):
         self.operations.append(AddData(values))
 
     @staticmethod
-    def _process_array(
-        values: Iterable[Any], schema: pt.array
-    ) -> Tuple[Callable[[Any], Any], bool]:
-        item_schema = schema.items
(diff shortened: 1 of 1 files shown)


graviti/dataframe/frame.py
```diff
@@ -33,6 +33,7 @@
 from graviti.utility import MAX_REPR_ROWS, File, PagingLists
 
 _T = TypeVar("_T", bound="DataFrame")
+_C = TypeVar("_C", bound="Container")
```
linjiX, line 36: it is better to paste the error message to the commit message body

linjiX (approved):

PR #238, opened 2022-05-13
ci: set the exact version for additional_dependencies in pre-commit

AChenQ:


.pre-commit-config.yaml
```diff
@@ -64,7 +64,14 @@ repos:
...
         exclude: "(test_\\w*\\.py$)"
         additional_dependencies:
-          [types-PyYAML, types-mock, types-requests, click, tensorbay, numpy>=1.22.0]
+          [
+            types-PyYAML==6.0.7,
+            types-requests==2.25.11,
+            tensorbay==1.20.0,
+            numpy==1.22.0,
+            types-mock,
+            click,
```
linjiX, line 73: why `types-mock` and `click` do not have version?
also do we need `click` here?

linjiX (changes requested):

.pre-commit-config.yaml
AChenQ: we don't need `click` and `types-mock` now, I will remove it in an another commit.

linjiX (approved):

PR #241, opened 2022-05-16
fix(manager): fix ArrowTypeError caused by wrong "_getter" in "Commit"

graczhual:


graviti/manager/commit.py
```diff
@@ -136,9 +136,23 @@ def search(self, sheet: str, criteria: Dict[str, Any]) -> DataFrame:
             The created :class:`~graviti.dataframe.DataFrame` instance.
 
         """
+        total_count = create_search(
```
Lee-000, line 139: Can we wrap `create_search` and common arguments into a method like `_create_search`?

linjiX (approved):

Lee-000 (approved):

PR #244, opened 2022-05-16
feat(portex): implement "read_json" and "read_yaml"

linjiX:


GravitiRob:
Task linked: [CU-2c1xuv8 添加 pt.read_yaml 与 pt.read_json 接口](https://app.clickup.com/t/2c1xuv8)

graviti/portex/base.py
```diff
@@ -255,3 +256,31 @@ def copy(self: _T) -> _T:
...
+
+
+def read_yaml(path: PathLike) -> PortexType:
+    """Read a yaml file into Portex type.
+
+    Arguments:
+        path: The path of the yaml file.
+
+    Returns:
+        A Portex type instance transferred from the input yaml file.
```
Lee-000, line 268: Is `transferred` accurate? Would `created` be better?

graviti/portex/base.py
```diff
@@ -255,3 +256,31 @@ def copy(self: _T) -> _T:
...
+
+    Arguments:
+        path: The path of the yaml file.
+
+    Returns:
+        A Portex type instance created from the input yaml file.
+
+    """
+    with open(path, encoding="utf-8") as fp:
+        return PortexType.from_pyobj(yaml.load(fp, yaml.Loader))
```
Lee-000, line 272: Why not use `from_yaml`?

Lee-000: Then should we change the input type of `from_yaml`? Since `yaml.load` accepts not only `str`.

linjiX: but json.loads not accept `fp` only `str`, I think we should change from_yaml and from_json together?

Lee-000 (approved):

PR #251, opened 2022-05-17
fix(portex): return "[]" in "get_keys" of template with Series container

Lee-000:


graviti/portex/template.py
```diff
@@ -83,6 +83,12 @@ def get_keys(self, type_name: Optional[str] = None) -> List[Tuple[str, ...]]:
             A list of keys to locate the data.
 
         """
+        container = EXTERNAL_TYPE_TO_CONTAINER.get(
```
linjiX, line 86: `self.container`?

linjiX (changes requested):

graviti/dataframe/column/series.py
```diff
@@ -53,6 +53,7 @@ class Series(Container):
 
     """
 
+    has_key = False
```
linjiX, line 56: `has_key` or `has_keys`?

linjiX (changes requested):

linjiX (approved):

AChenQ (approved):

PR #261, opened 2022-05-18
refactor(portex): adapt the JSON array grammar of Portex template params

Lee-000:


graviti/portex/template.py
```diff
@@ -190,7 +190,7 @@ def template(
     type_: Type[PortexExternalType] = type(
         name,
         (PortexExternalType,),
-        {"parameters": params, "dependences": factory.dependences, "factory": factory},
+        {"params": params, "dependences": factory.dependences, "factory": factory},
```
linjiX, line 193: what is this?

linjiX (changes requested):

graviti/portex/template.py
Lee-000: This is the attr of `PortexExternalType`, which should not be the key `parameters` in yaml.

graviti/portex/param.py
```diff
@@ -169,19 +167,19 @@ def __init__(self, values: Optional[Mapping[str, Param]] = None) -> None:
         self._data: MutableMapping[str, Param] = OrderedDict(values if values else {})
 
     @classmethod
-    def from_pyobj(cls, pyobj: Dict[str, Any]) -> "Params":
-        """Create Params instance from python dict.
+    def from_pyobj(cls, pyobj: List[Dict[str, Any]]) -> "Params":
```
linjiX, line 170: shall we also update the `Params.to_pyobj`?

linjiX (changes requested):

linjiX (approved):

PR #265, opened 2022-05-20
docs(openapi): correct the function name of Examples in "data.py"

graczhual:


graviti/openapi/data.py
```diff
@@ -255,7 +255,7 @@ def update_data(
         data: The update data.
 
     Examples:
-        >>> update_dataset(
+        >>> update_data(
```
linjiX, line 258: commit message is not clear, should we mention "correct the function name in example"?

linjiX (changes requested):

linjiX (approved):

PR #270, opened 2022-05-23
refactor: modify the default value of "offset" and "limit"

graczhual:
offset in openapi: 0 -> None
limit in openapi: 128 -> None
limit in manager(except for list data): 128 -> 24

### DIFF
--- a/graviti/manager/branch.py
+++ b/graviti/manager/branch.py
@@ -55,7 +55,7 @@ class BranchManager:
     def __init__(self, dataset: "Dataset") -> None:
         self._dataset = dataset
 
-    def _generate(self, offset: int = 0, limit: int = 128) -> Generator[Branch, None, int]:
+    def _generate(self, offset: int, limit: int) -> Generator[Branch, None, int]:
         response = list_branches(
             self._dataset.access_key,
             self._dataset.url,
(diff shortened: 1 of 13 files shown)


graviti/manager/branch.py
```diff
@@ -55,7 +55,9 @@ class BranchManager:
     def __init__(self, dataset: "Dataset") -> None:
         self._dataset = dataset
 
-    def _generate(self, offset: int = 0, limit: int = 128) -> Generator[Branch, None, int]:
+    def _generate(
+        self, offset: Optional[int] = None, limit: Optional[int] = None
```
linjiX, line 59: shall we add default for `_generate` function?
will it better to remove the default value for `_generate`?

linjiX (changes requested):

graviti/manager/branch.py
```diff
@@ -55,7 +55,9 @@ class BranchManager:
     def __init__(self, dataset: "Dataset") -> None:
         self._dataset = dataset
 
-    def _generate(self, offset: int = 0, limit: int = 128) -> Generator[Branch, None, int]:
+    def _generate(
+        self, offset: Optional[int], limit: Optional[int]
```
linjiX, line 59: `_generate` is only for `LazyPagingList`, I think the offset and limit type can be just `int`

linjiX (changes requested):

linjiX (approved):

PR #284, opened 2022-05-26
fix(portex): use AvroFixedArraySchema to convert Portex tensor

Lee-000:


graviti/portex/avro.py
```diff
@@ -208,7 +208,7 @@ def _on_fixed_shape_list(names, namespace, name, _pa_ist: tensor) -> AvroArraySc
     sub_namespace = f"{namespace}.{name}.items"
     sub_name = "items"
     items = _on_type(names, sub_namespace, sub_name, 0, _pa_ist.items.to_builtin())
-    return AvroArraySchema(items=items, shape=_pa_ist.shape)
+    return AvroFixedArraySchema(items=items, shape=_pa_ist.shape)
```
linjiX, line 211: should the return type be updated?

linjiX (approved):

PR #286, opened 2022-05-26
refactor: modify the url predix of OpenAPIv2

graczhual:
"https://api.graviti.com" ->
"https://gas.graviti.com/gatewayv2/tensorbay-open-api"

### DIFF
--- a/graviti/openapi/user.py
+++ b/graviti/openapi/user.py
@@ -21,7 +21,10 @@ def get_current_user(access_key: str, url: str) -> Dict[str, Any]:
         The response of OpenAPI.
 
     Examples:
-        >>> get_current_user("ACCESSKEY-********", "https://api.graviti.com")
+        >>> get_current_user(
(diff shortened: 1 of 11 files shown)


graviti/openapi/user.py
```diff
@@ -21,7 +21,10 @@ def get_current_user(access_key: str, url: str) -> Dict[str, Any]:
         The response of OpenAPI.
 
     Examples:
-        >>> get_current_user("ACCESSKEY-********", "https://api.graviti.com")
+        >>> get_current_user(
+        ...     "ACCESSKEY-********",
+        ...     "https://gas.graviti.com/gatewayv2/tensorbay-open-api",
```
linjiX, line 26: may be we can keep the url in example unchanged?

graczhual: But this example is wrong, although few users will see it here.

linjiX (approved):

PR #290, opened 2022-05-27
refactor(manager): use "get_revision" to replace "if-else" in "checkout"

graczhual:


graviti/manager/dataset.py
```diff
@@ -11,20 +11,27 @@
...
     create_dataset,
     delete_dataset,
     get_dataset,
+    get_revision,
     list_datasets,
     update_dataset,
 )
 from graviti.utility import ReprMixin, ReprType, UserMutableMapping
 
+_REVISION_TYPES = {
```
linjiX, line 29: will it be better to use python `enum.Enum`?

linjiX (changes requested):

graviti/manager/dataset.py
graczhual: It has been modified. Should we use `enum` instead of `dict` in similar scenarios? What are the benefits?

linjiX (approved):

PR #293, opened 2022-05-27
refactor(dataframe): comment out unimplemented methods in "DataFrame"

AChenQ:


graviti/dataframe/frame.py
```diff
@@ -377,19 +377,19 @@ def shape(self) -> Tuple[int, int]:
...
+    #
+    #     Return:
+    #         Size of the DataFrame.
+    #
+    #     Examples:
+    #         >>> df = DataFrame({"col1": [1, 2], "col2": [3, 4]})
+    #         >>> df.size
+    #         4
+    #
+    #     """
```
linjiX, line 392: normally we use word `method` to represent the function in class,
so better to use `method` in commit message

linjiX (changes requested):

graviti/dataframe/frame.py
linjiX: Also this commit will affect the SDK API, use `refactor` is better.

linjiX (changes requested):

linjiX (approved):

PR #294, opened 2022-05-28
fix(portex): fix "PortexType.imports" does not add itself into "Imports"

linjiX:


GravitiRob:
Task linked: [CU-27k0kcm 修复 PortexType.imports 不包含当前类型的问题](https://app.clickup.com/t/27k0kcm)

PR #299, opened 2022-05-30
feat(manager): support creating draft on the current branch by default

graczhual:


graviti/manager/draft.py
```diff
@@ -267,7 +269,16 @@ def create(
...
             The :class:`.Draft` instance with the given title and description.
 
+        Raises:
+            StatusError: When creating the draft without basing on a branch.
+
         """
+        if branch is None:
+            current_branch = self._dataset.HEAD
+            if not isinstance(current_branch, Branch):
+                raise StatusError("Creating the draft without basing on a branch is not allowed")
```
linjiX, line 279: shall we mention the dataset is not on a branch in the commit message?

linjiX: shall we mention the dataset is not on a branch in the error message?

linjiX (changes requested):

graviti/manager/draft.py
```diff
@@ -267,7 +269,19 @@ def create(
...
+        Raises:
+            StatusError: When creating the draft without basing on a branch.
+
         """
+        if branch is None:
+            current_branch = self._dataset.HEAD
+            if not isinstance(current_branch, Branch):
+                raise StatusError(
+                    "The current dataset is not on a branch, please checkout target branch \
+first or input the argument 'branch'"
```
linjiX, line 281: if a string is too long, you just use two strings without `,`, like:
```python
raise StatusError(
    "The current dataset is not on a branch, please checkout target branch "
    "first or input the argument 'branch'"
```

linjiX: also we do not have the concept "target branch" now, maybe "checkout a branch"?

linjiX (changes requested):

linjiX (approved):

PR #302, opened 2022-05-30
refactor(manager): add check for operations when commit id is None

Lee-000:


graviti/manager/branch.py
```diff
@@ -79,12 +79,19 @@ def create(self, name: str, revision: Optional[str] = None) -> Branch:
...
+
         Returns:
             The :class:`.Branch` instance with the given name.
 
         """
         if not revision:
             revision = self._dataset.HEAD.commit_id
+            if revision is None:
+                raise ForbiddenError(
+                    "Can not create tag on the default branch without commit history"
```
linjiX, line 93: I think the word `Can not` is not clear enough.
Will it be better if we use `xxx is not allowed` or `xxx is forbidden` to point out that this behavior is forbidden by design not other reason?

graviti/manager/commit.py
```diff
@@ -84,6 +85,11 @@ def _list_data(self, offset: int, limit: int, sheet_name: str) -> Dict[str, Any]
         )["data"]
 
     def _list_sheets(self) -> Dict[str, Any]:
+        if self.commit_id is None:
+            raise StatusError(
+                "No commit on the current branch. Please create a draft and add data first"
```
linjiX, line 90: why we need add data? can the user just create an empty draft?

linjiX (changes requested):

graviti/manager/commit.py
```diff
@@ -267,12 +280,17 @@ def get(self, revision: Optional[str] = None) -> Commit:
             revision: The information to locate the specific commit, which can be the commit id,
```
graczhual, line 280: do we need to add if-else to `_generate commits`

Lee-000: If the commit id is None, the revision won't be added to the params of the request.

linjiX (approved):

graczhual (approved):

PR #316, opened 2022-06-02
refactor(exception): use decorator to register ResponseError

Lee-000:


GravitiRob:
Task linked: [CU-2juqx00 使用装饰器或者 "__init_subclass__" 重构 ResponseError 的分发功能](https://app.clickup.com/t/2juqx00)

graviti/exception.py
```diff
@@ -64,6 +66,7 @@ class ResponseError(ClientError):
     _INDENT = " " * len(__qualname__)  # type: ignore[name-defined]
 
     STATUS_CODE: int
+    MESSAGE_CODE: str
```
linjiX, line 69: `MESSAGE_CODE`? do we have a better name?

graviti/exception.py
```diff
@@ -74,6 +77,7 @@ def __init__(
 
     def __init_subclass__(cls) -> None:
         cls._INDENT = " " * len(cls.__name__)
+        RESPONSE_ERROR_DISTRIBUTOR[cls.MESSAGE_CODE] = cls
```
linjiX, line 80: I think use decorator to register the errors into distributor is easier to understand

graviti/exception.py
```diff
@@ -129,18 +129,53 @@ def __str__(self) -> str:
...
+class ResponseErrorRegister:
+    """A class decorator to register the ResponseError into the distributor.
+
+    Arguments:
+        error_code: The response error code of the specific ResponseError.
+
+    """
+
+    def __init__(self, error_code: str) -> None:
+        self._error_code = error_code
```
linjiX, line 147: error_code? do we have a better name?
for me code should be int not str.

graviti/exception.py
```diff
@@ -129,18 +129,53 @@ def __str__(self) -> str:
...
+        Returns:
+            The input response error class unchanged.
+
+        """
+        RESPONSE_ERROR_DISTRIBUTOR[self._error_code] = response_error
+
+        return response_error
+
+
+@ResponseErrorRegister("AccessDenied")
```
linjiX, line 164: Shall we put `STATUS_CODE` into the decorator?

linjiX (changes requested):

graviti/exception.py
Lee-000: We do not use `STATUS_CODE` to identify the errors currently, so I don't think it is necessary to put it into the decorator.

linjiX (approved):

PR #318, opened 2022-06-02
feat(portex): raise "GitNotFoundError" when "git" command not found

linjiX:


GravitiRob:
Task linked: [CU-2f8wzj6 对 builder 中的 git 命令添加详细的报错信息](https://app.clickup.com/t/2f8wzj6)

graviti/exception.py
```diff
@@ -27,6 +27,52 @@ def __str__(self) -> str:
...
+class PortexError(GravitiException):
+    """This is the base class for custom exceptions in Graviti portex module."""
+
+
+class GitNotFoundError(PortexError):
+    """This class defines the exception for the git command not found error.
+
+    Arguments:
+       message: The error message.
+       file_not_found_error: The FileNotFoundError raised from the subprocess.run().
```
Lee-000, line 39: What is this?

graviti/exception.py
```diff
@@ -27,6 +27,52 @@ def __str__(self) -> str:
...
+
+    """
+
+    _MESSAGE = (
+        "'git' command failed, most likely due to the 'git' is not installed.\n"
+        "Please install 'git' first: https://git-scm.com/book/en/v2/Getting-Started-Installing-Git"
+    )
+
+    def __init__(self, message: str = _MESSAGE):
```
Lee-000, line 48: Should we add a class attr default message to `GravitiException` and rewrite its init method?

graviti/exception.py
linjiX: removed

graviti/exception.py
linjiX: actually I think we can remove the `message` in `GravitiException`

linjiX: Maybe we can add a `MessageMixin` to handle message.

Lee-000 (approved):

PR #328, opened 2022-06-09
feat(manager): add "check_head_status" to some methods in "manager"

graczhual:
- create branch
- list branches
- get branch
- create draft
- list commits
- get commit
- create tag
- list tags
- get tag

### DIFF
--- a/graviti/manager/branch.py
+++ b/graviti/manager/branch.py
@@ -12,6 +12,7 @@ from tensorbay.utility import attr
 
 from graviti.exception import NoCommitsError, ResourceNameError
 from graviti.manager.commit import NamedCommit
+from graviti.manager.common import check_head_status
 from graviti.manager.lazy import LazyPagingList
 from graviti.openapi import create_branch, delete_branch, get_branch, list_branches
 
@@ -65,7 +66,9 @@ class BranchManager:
             limit=limit,
         )
 
+        head = self._dataset.HEAD
         for item in response["branches"]:
+            check_head_status(head, item["name"], item["commit_id"])
             yield Branch(self._dataset, **item)
 
         return response["total_count"]  # type: ignore[no-any-return]
@@ -86,8 +89,9 @@ class BranchManager:
             The :class:`.Branch` instance with the given name.
 
         """
+        head = self._dataset.HEAD
         if not revision:
-            revision = self._dataset.HEAD.commit_id
+            revision = head.commit_id
             if revision is None:
                 raise NoCommitsError(
                     "Creating branches on the default branch without commit history is not allowed."
@@ -102,6 +106,8 @@ class BranchManager:
             name=name,
             revision=revision,
         )
+
+        check_head_status(head, revision, response["commit_id"])
         return Branch(self._dataset, **response)
 
     def get(self, name: str) -> Branch:
@@ -127,6 +133,10 @@ class BranchManager:
             self._dataset.name,
             branch=name,
         )
+
+        if getattr(self._dataset, "HEAD", None):
+            check_head_status(self._dataset.HEAD, name, response["commit_id"])
+
         return Branch(self._dataset, **response)
 
     def list(self) -> LazyPagingList[Branch]:
--- a/graviti/manager/commit.py
+++ b/graviti/manager/commit.py
@@ -11,6 +11,7 @@ from tensorbay.utility import AttrsMixin, attr
 
 from graviti.dataframe import DataFrame
 from graviti.exception import NoCommitsError
+from graviti.manager.common import check_head_status
 from graviti.manager.lazy import LazyPagingList
 from graviti.manager.sheets import Sheets
 from graviti.openapi import (
@@ -248,11 +249,8 @@ class CommitManager:
     def __init__(self, dataset: "Dataset") -> None:
         self._dataset = dataset
 
-    def _generate(
-        self, revision: Optional[str], offset: int, limit: int
-    ) -> Generator[Commit, None, int]:
-        if revision is None:
-            revision = self._dataset.HEAD.commit_id
+    def _generate(self, revision: str, offset: int, limit: int) -> Generator[Commit, None, int]:
+        head = self._dataset.HEAD
 
         response = list_commits(
             self._dataset.access_key,
@@ -264,7 +262,11 @@ class CommitManager:
             limit=limit,
         )
 
-        for item in response["commits"]:
+        commits = response["commits"]
+        if offset == 0 and commits:
+            check_head_status(head, revision, commits[0]["commit_id"])
+
+        for item in commits:
             yield Commit(self._dataset, **item)
 
         return response["total_count"]  # type: ignore[no-any-return]
@@ -284,8 +286,9 @@ class CommitManager:
             The :class:`.Commit` instance with the given revision.
 
         """
+        head = self._dataset.HEAD
         if revision is None:
-            revision = self._dataset.HEAD.commit_id
+            revision = head.commit_id
             if revision is None:
                 raise NoCommitsError("No commits on the default branch yet")
 
@@ -300,6 +303,8 @@ class CommitManager:
             raise NoCommitsError("No commits on the default branch yet")
 
         del response["type"]
+        check_head_status(head, revision, response["commit_id"])
+
         return Commit(self._dataset, **response)
 
     def list(self, revision: Optional[str] = None) -> LazyPagingList[Commit]:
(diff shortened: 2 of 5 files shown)


GravitiRob:
Task linked: [CU-2fzzmyr 为版本管理相关接口添加 branch 不同步检查，当发现不同步时，打印警告信息](https://app.clickup.com/t/2fzzmyr)

graviti/manager/tag.py
```diff
@@ -74,8 +75,11 @@ def create(self, name: str, revision: Optional[str] = None) -> Tag:
             The :class:`.Tag` instance with the given name.
 
         """
+        dataset_head = self._dataset.HEAD
+        input_revision = True
```
linjiX, line 79: `input_revision` seems useless.

graviti/manager/tag.py
```diff
@@ -74,8 +75,11 @@ def create(self, name: str, revision: Optional[str] = None) -> Tag:
             The :class:`.Tag` instance with the given name.
 
         """
+        dataset_head = self._dataset.HEAD
```
linjiX, line 78: i think `head` is ok

graviti/manager/commit.py
```diff
@@ -264,7 +269,18 @@ def _generate(
             limit=limit,
         )
 
-        for item in response["commits"]:
+        commits = response["commits"]
+        if (
+            input_revision
+            and commits
+            and getattr(dataset_head, "name", None) == revision
+            and dataset_head.commit_id != commits[0]["commit_id"]
```
linjiX, line 277: because `list_commits` has `offset`, `commits[0]` may not be the first commit.

linjiX (changes requested):

graviti/manager/branch.py
```diff
@@ -102,6 +104,10 @@ def create(self, name: str, revision: Optional[str] = None) -> Branch:
             name=name,
             revision=revision,
         )
+
+        if getattr(head, "name", None) == revision and head.commit_id != response["commit_id"]:
+            warnings.warn(StatusWarning(type(head).__name__.lower(), revision), stacklevel=2)
```
linjiX, line 109: `head.__class__.__name__`?

graviti/manager/branch.py
```diff
@@ -102,6 +104,10 @@ def create(self, name: str, revision: Optional[str] = None) -> Branch:
...
+
         return Branch(self._dataset, **response)
 
     def get(self, name: str) -> Branch:
```
linjiX, line 113: do we need a default value for the `name` in `BranchManager.get`

graviti/manager/branch.py
```diff
@@ -127,6 +133,14 @@ def get(self, name: str) -> Branch:
             self._dataset.name,
             branch=name,
         )
+
+        if (
+            hasattr(self._dataset, "HEAD")
```
linjiX, line 138: In what condition, the `self._dataset` does not have `"HEAD"`

graviti/manager/commit.py
```diff
@@ -251,8 +252,9 @@ def __init__(self, dataset: "Dataset") -> None:
     def _generate(
         self, revision: Optional[str], offset: int, limit: int
```
linjiX, line 253: shall we set the type of `revision` to `str` in `_generate` rather than `Optional[str]`

graviti/manager/draft.py
```diff
@@ -300,6 +301,9 @@ def create(
             branch=branch,
             description=description,
         )
+
+        if getattr(head, "name", None) == branch and head.commit_id != response["parent_commit_id"]:
```
linjiX, line 305: Is `head` must be `Branch` here? do we need to use `getattr`?

linjiX (changes requested):

graviti/manager/branch.py
graczhual: initialize the dataset in `ws.datasets.get()` and `ws.datasets.list()`

graviti/manager/commit.py
graczhual: move the `revision = head.commit_id` to `list()` instead of `_generate()`? But here the revision could be None, which means the default commit

graviti/manager/draft.py
graczhual: head could be commit.
```
dataset.checkout("commit_id")
dataset.drafts.create("draft-1", branch="main")
```

graviti/manager/branch.py
graczhual: 1. Get the default branch
2. If the HEAD is branch, get the current branch. Otherwise, the `name` is required. In my opinion, this is better.

graviti/manager/tag.py
```diff
@@ -115,6 +122,10 @@ def get(self, name: str) -> Tag:
             self._dataset.name,
             tag=name,
         )
+
+        if getattr(head, "name", None) == name and head.commit_id != response["commit_id"]:
```
linjiX, line 126: can we wrap these two lines as a function?

graviti/manager/branch.py
linjiX: Can we make OpenAPI list/get datasets return the default branch details

linjiX (changes requested):

graviti/manager/branch.py
graczhual: The backend also needs to call "get branch" again. So I thought it was not necessary to change.

graviti/manager/branch.py
```diff
@@ -127,6 +131,10 @@ def get(self, name: str) -> Branch:
             self._dataset.name,
             branch=name,
         )
+
+        if hasattr(self._dataset, "HEAD"):
```
Lee-000, line 135: Why check whether the dataset has `HEAD` here?

graviti/manager/common.py
```diff
@@ -0,0 +1,27 @@
...
+import warnings
+from typing import TYPE_CHECKING
+
+from graviti.exception import StatusWarning
+
+if TYPE_CHECKING:
+    from graviti.manager.commit import Commit
+
+
+def check_dataset_status(head: "Commit", revision_name: str, revision_commit_id: str) -> None:
```
linjiX, line 17: Is `check_head_status` better?

linjiX: `revision_name` -> `remote_revision`
`revision_commit_id` -> `remote_commit_id`

linjiX (changes requested):

graviti/exception.py
```diff
@@ -11,6 +11,27 @@
 from requests.models import Response
 
 
+class StatusWarning(Warning):
```
linjiX, line 14: move this to `manager/common.py`?

linjiX (changes requested):

graviti/manager/branch.py
graczhual: dataset has no attribute HEAD when we initialize it.

graviti/manager/branch.py
```diff
@@ -127,6 +132,10 @@ def get(self, name: str) -> Branch:
...
```
linjiX, line 136: `getattr` is faster than `hasattr`, especially when the attr is a property

graviti/manager/branch.py
```diff
@@ -66,6 +67,7 @@ def _generate(self, offset: int, limit: int) -> Generator[Branch, None, int]:
         )
 
         for item in response["branches"]:
+            check_head_status(self._dataset.HEAD, item["name"], item["commit_id"])
```
linjiX, line 70: set `head = self._dataset.HEAD` to avoid the `.` operation in loop?

graviti/manager/common.py
```diff
@@ -0,0 +1,46 @@
...
+
+    Arguments:
+       revision_type: The type of current dataset revision.
+       name: The name of current dataset revision.
+
+    """
+
+    def __init__(self, revision_type: str, name: str):
+        super().__init__()
+        self._type = revision_type
```
linjiX, line 26: we have a enum for revision_type, shall we use it here?

linjiX (changes requested):

Lee-000 (approved):

graviti/manager/common.py
graczhual: we can not import `RevisionType` in `common.py`, this will cause a `circular import`. And we can not move the `RevisionType` to `common.py`. 
RevisionType -> need Branch -> need check_head_status -> need RevisionType

graviti/manager/common.py
```diff
@@ -0,0 +1,46 @@
...
+        self._name = name
+
+    def __str__(self) -> str:
+        return (
+            f"The commit id of {self._type} ({self._name}) of current dataset is not up-to-date. It"
+            f" is recommended to update the 'dataset.HEAD' by the method 'checkout('{self._name}')'"
```
linjiX, line 32: the variable name of current dataset may not be `dataset`, I think we should not use `dataset.HEAD`

linjiX (approved):

PR #334, opened 2022-06-10
feat(portex): add ImmuatbleFields and MutableFields

Lee-000:


GravitiRob:
Task linked: [CU-2g5p6ph 增加 FixedFields 类型表示无法修改的 fields](https://app.clickup.com/t/2g5p6ph)

graviti/portex/field.py
```diff
@@ -11,11 +11,95 @@
...
+        self._check_key(name)
+        self._check_value(portex_type)
+
+        if self.__contains__(name):
+            raise KeyError(f'"{name}" already exists in the Fields')
+
+        self._keys.insert(index, name)
+        self._data.__setitem__(name, portex_type)
+
+    def astype(self, name: str, portex_type: PortexType) -> None:
```
linjiX, line 71: what if user use `__setitem__` to change type of fields?

graviti/portex/field.py
```diff
@@ -11,11 +11,95 @@
...
+
+    def __repr__(self) -> str:
+        return self._repr1(0)
+
+    @classmethod
+    def _check_value(cls, value: PortexType) -> None:
+        if not isinstance(value, PortexType):
+            raise TypeError(f'The value in "{cls.__name__}" should be a PortexType')
+
+    def _repr1(self, level: int) -> str:
```
linjiX, line 30: what is the difference between the `__repr__` in `InmutableFields` and normal `Fields`?

graviti/utility/collections.py
```diff
@@ -430,6 +413,32 @@ def _setitem(self, key: str, value: _V) -> None:
             self._keys.append(key)
         self._data.__setitem__(key, value)
 
+
+class NameOrderedDict(MutableMapping[str, _V], InmutableNameOrderedDict[_V]):
```
linjiX, line 417: in python builtin collections library, the mutable class will have `Mutable` prefix, like `MutableMapping`, and the Immutable class will have no prefix, like `Mapping`. I think we should follow this convention.

linjiX (changes requested):

graviti/portex/field.py
Lee-000: It is also allowed.

graviti/utility/collections.py
linjiX: Here `Dict` means `Mutable`, so that in collections we have `UserDict`, `OrderedDict` they are all mutable,
here there is another naming convention for `immutable dict` which is `frozen dict`, so I think we can use `FrozenNameOrderedDict` for what you need.

graviti/portex/field.py
Lee-000: No difference.

graviti/utility/collections.py
linjiX: here is some reference:
https://docs.python.org/zh-cn/3/library/stdtypes.html?highlight=frozenset#frozenset
https://peps.python.org/pep-0416/

graviti/portex/field.py
Lee-000: `Fields` will be replaced with `ConnectedFields` in the future, do we need to reuse this part of logic now?

linjiX (approved):

graczhual (approved):

PR #342, opened 2022-06-14
refactor(manager): add class "DefaultValue" to "common.py"

graczhual:
To make it easier for the user to understand, use class "DefaultValue" instead
of None to represent the default value of parameters from methods in manager.

### DIFF
--- a/graviti/manager/common.py
+++ b/graviti/manager/common.py
@@ -6,10 +6,30 @@
 """Common tools."""
 
 import warnings
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 if TYPE_CHECKING:
     from graviti.manager.commit import Commit
+
+
+class DefaultValue:
+    """This class defines the default value of parameters from methods in manager.
+
+    Arguments:
+        name: The name of the default value.
+
+    """
+
+    def __init__(self, name: str) -> None:
+        self._name = name
+
+    def __repr__(self) -> str:
+        return self._name
+
+
+CURRENT_COMMIT: Any = DefaultValue("CURRENT_COMMIT")
+CURRENT_BRANCH: Any = DefaultValue("CURRENT_BRANCH")
+ALL_BRANCHES: Any = DefaultValue("ALL_BRANCHES")
 
 
 class StatusWarning(Warning):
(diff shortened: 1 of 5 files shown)


GravitiRob:
Task linked: [CU-28v2h00 SDK 版本管理相关函数默认值可读性优化](https://app.clickup.com/t/28v2h00)

graviti/manager/common.py
```diff
@@ -6,12 +6,20 @@
...
 
 if TYPE_CHECKING:
     from graviti.manager.commit import Commit
 
 
+class DatasetStatus(Enum):
+    """An enumeration type representing the current status of the dataset."""
+
+    CURRENT_COMMIT = object()
+    CURRENT_BRANCH = object()
```
linjiX, line 20: normally we use `auto()` to generate enum values automatically.
https://docs.python.org/zh-cn/3/library/enum.html#enum.auto

graviti/manager/dataset.py
```diff
@@ -112,8 +110,6 @@ def __init__(
         self.is_public = is_public
         self.config = config
         self._data: Commit = self.branches.get(default_branch)
-        if commit_id != self._data.commit_id:
-            self._data = self.commits.get(commit_id)
```
linjiX, line 116: shall we use another PR for removing `Dataset.commit_id`?

linjiX (changes requested):

graviti/manager/common.py
```diff
@@ -6,12 +6,32 @@
...
 
 import warnings
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 if TYPE_CHECKING:
     from graviti.manager.commit import Commit
 
 
+class ArgumentsDefaultValue:
```
linjiX, line 15: I think `DefaultValue` is ok.

graviti/manager/draft.py
```diff
@@ -250,11 +250,14 @@ def __init__(self, dataset: "Dataset") -> None:
 
     def _generate(
         self,
-        state: Optional[str],
+        state: str,
         branch: Optional[str],
```
linjiX, line 254: do we need to change the type of `branch` to `str`?

graczhual: but we need to set `branch = None` here

linjiX (changes requested):

linjiX (approved):

PR #361, opened 2022-06-16
fix(dataframe): fix schema loss error after DataFrame.__getitem__

AChenQ:


graviti/dataframe/frame.py
```diff
@@ -128,7 +128,9 @@ def __getitem__(self, key: Union[str, Iterable[str]]) -> Container:
 
         if isinstance(key, Iterable):
             new_columns = {name: self._columns[name] for name in key}
-            return self._construct(new_columns)
+            fields = self.schema.to_builtin().fields  # type: ignore[attr-defined]
+            schema = pt.record({name: fields[name] for name in new_columns.keys()})
```
linjiX, line 132: `new_columns.keys()` -> `new_columns`

linjiX (changes requested):

graviti/dataframe/frame.py
linjiX: will it be better to get the schema from the columns?
```python
schema = pt.record({name: column.schema for name, column in new_columns.items()})
```

linjiX (changes requested):

linjiX (approved):

PR #362, opened 2022-06-16
feat: add "_to_patch_data" to "DataFrame" to support updating data

graczhual:


GravitiRob:
Task linked: [CU-2h0za7f 实现支持更新数据的 Operation](https://app.clickup.com/t/2h0za7f)

graviti/dataframe/frame.py
```diff
@@ -303,6 +311,17 @@ def _extend(self, values: "DataFrame") -> None:
         for key, value in self._columns.items():
             value._extend(values[key])  # pylint: disable=protected-access
 
+    def _to_post_data(self) -> List[Dict[str, Any]]:
+        names = self._column_names.copy()
+        values = list(self._columns.values())
+
+        names.append(RECORD_KEY)
+        values.append(self._record_key.to_pyarrow())  # type: ignore[union-attr]
```
linjiX, line 319: shall we change the type of `_record_key` to ColumnSeries?

linjiX (changes requested):

graviti/dataframe/frame.py
graczhual: a special ColumnSeries without schema?

linjiX: _record_key is string, it can have schema pt.string()

graviti/dataframe/frame.py
```diff
@@ -627,12 +649,13 @@ def extend(self, values: Union[Iterable[Dict[str, Any]], "DataFrame"]) -> None:
         elif self.operations is not None:
             values = values.copy()
 
+        values_length = len(values)
```
linjiX, line 652: why we need `values_length`?

linjiX (changes requested):

graviti/dataframe/frame.py
graczhual: if we use extend like `df.extend(df)`, then the `len(values)` will double

linjiX (approved):

AChenQ (approved):

PR #374, opened 2022-06-20
refactor(manager): use "count criteria" to get total count in search

AChenQ:


graviti/manager/commit.py
```diff
@@ -162,7 +163,17 @@ def _getter(offset: int, limit: int) -> Dict[str, Any]:
                 limit=limit,
             )
 
-        total_count = _getter(0, 1)["total_count"]
+        count_criteria = deepcopy(criteria)
+        count_criteria["select"] = [{"count": {"$count": ["$.*"]}}]
```
graczhual, line 167: do we need to set limit = 1 here?

graviti/manager/commit.py
```diff
@@ -162,7 +163,17 @@ def _getter(offset: int, limit: int) -> Dict[str, Any]:
                 limit=limit,
             )
 
-        total_count = _getter(0, 1)["total_count"]
+        count_criteria = deepcopy(criteria)
```
linjiX, line 166: why we need `deepcopy` here?

linjiX (changes requested):

graviti/manager/commit.py
AChenQ: Dont't need limit here, just the count will be returned.

graviti/manager/commit.py
AChenQ: To prevent the `criteria` from changing.

graczhual (approved):

linjiX (approved):

PR #384, opened 2022-06-21
refactor(portex): rename "ImmutableFields" and remove "MutableFields"

Lee-000:
1. Rename "ImmutableFields" to "FrozenFields".
2. Remove "MutableFields".

### DIFF
--- a/graviti/portex/field.py
+++ b/graviti/portex/field.py
@@ -32,8 +32,8 @@ from graviti.utility import FrozenNameOrderedDict, NameOrderedDict
 _T = TypeVar("_T", bound="Fields")
 
 
-class ImmutableFields(FrozenNameOrderedDict[PortexType]):
-    """Represents an immutable fields dict."""
+class FrozenFields(FrozenNameOrderedDict[PortexType]):
+    """Represents a frozen fields dict."""
 
     @classmethod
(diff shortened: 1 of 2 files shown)


GravitiRob:
Task linked: [CU-2hkcfbq 删除 MutableFields，将 ImmutableFields 改名为 FrozenFields](https://app.clickup.com/t/2hkcfbq)

graviti/portex/field.py
```diff
@@ -283,7 +227,7 @@ def rename(self, old_name: str, new_name: str) -> None:
         self._get_fields(old_name).rename(old_name, new_name)
 
 
-class Fields(NameOrderedDict[PortexType]):
+class Fields(NameOrderedDict[PortexType], FrozenFields):  # type: ignore[misc]
```
linjiX, line 230: where does the mypy misc comes from？

linjiX (changes requested):

graviti/portex/field.py
Lee-000: Mypy considers that definition of "__setitem__" in base class "MutableMapping" is incompatible with the definition in base class "FrozenFields".

linjiX (approved):

PR #389, opened 2022-06-21
refactor(portex): make "container" in "PortexType" to "ClassVar"

linjiX:
| cases                  | before              | after              |
| ---------------------- | ------------------- | ------------------ |
| std.label.Box2D()      | 135 µs ± 279 ns     | 27.1 µs ± 244 ns   |
| std.label.Box3D()      | 280 µs ± 4.56 µs    | 27.1 µs ± 298 ns   |

### DIFF
--- a/graviti/portex/external.py
+++ b/graviti/portex/external.py
@@ -34,9 +34,14 @@ class PortexExternalType(PortexType):  # pylint: disable=abstract-method
 
         container = EXTERNAL_TYPE_TO_CONTAINER.get(
             (cls.package.url, cls.package.revision, cls.__name__)
-        )
+        ) or getattr(cls.factory.class_, "container", None)
+
         if container:
             cls.container = container
+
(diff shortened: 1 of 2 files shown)


graviti/portex/external.py
```diff
@@ -35,9 +35,16 @@ def __init_subclass__(cls) -> None:
         container = EXTERNAL_TYPE_TO_CONTAINER.get(
             (cls.package.url, cls.package.revision, cls.__name__)
         )
+        if not container:
+            container = getattr(cls.factory.class_, "container", None)
+
```
Lee-000, line 40: `container = EXTERNAL_TYPE_TO_CONTAINER.get(
            (cls.package.url, cls.package.revision, cls.__name__)
        ) or getattr(cls.factory.class_, "container", None)`
        
What about using `or` here?

Lee-000 (approved):

PR #390, opened 2022-06-22
fix(dataframe): use to_builtin() when calling DataFrame._process_array

AChenQ:


graviti/dataframe/frame.py
```diff
@@ -682,7 +682,9 @@ def _get_process(value: Any, schema: pt.PortexType) -> Tuple[Callable[[Any], Any
         if container == DataFrame:
             return DataFrame._process_record(value, schema)  # type: ignore[arg-type]
         if container == ArraySeries:
-            return DataFrame._process_array(value, schema.items)  # type: ignore[attr-defined]
+            return DataFrame._process_array(
+                value, schema.to_builtin().items  # type: ignore[attr-defined]
```
linjiX, line 686: the commit type is fix, but I can not get what you fixed from the commit message.

linjiX: shall we mention the exception type in commit message?

linjiX (approved):

PR #393, opened 2022-06-22
fix(portex): fix the incorrect "abc" module of PortexRecordBase subclass

Lee-000:
reference: https://stackoverflow.com/questions/49457441/subclass-module \
        -set-to-metaclass-module-when-manually-creating-new-class-wit


GravitiRob:
Task linked: [CU-2j4g7c5 PortexRecordBase + PortexExternalType 的子类的 __module__ 特性变为 "abc"](https://app.clickup.com/t/2j4g7c5)

PR #394, opened 2022-06-22
refactor(dataframe): replace "_column_names" with "schema" in DataFrame

Lee-000:


GravitiRob:
Task linked: [CU-2ht8241 使用 Schema 替代 DataFrame._column_names](https://app.clickup.com/t/2ht8241)

graviti/dataframe/frame.py
```diff
@@ -165,7 +164,7 @@ def __repr__(self) -> str:
         return "\n".join(lines)
 
     def __len__(self) -> int:
-        return self._columns[self._column_names[0]].__len__()
+        return self._columns[list(self.schema.keys())[0]].__len__()
```
linjiX, line 167: do we need to use schema here?

linjiX (changes requested):

graviti/dataframe/frame.py
```diff
@@ -165,7 +164,7 @@ def __repr__(self) -> str:
         return "\n".join(lines)
 
     def __len__(self) -> int:
-        return self._columns[self._column_names[0]].__len__()
+        return list(self._columns.values())[0].__len__()
```
linjiX, line 167: will it be better to use `next` rather than `list`?

linjiX (changes requested):

PR #398, opened 2022-06-24
perf(portex): use dict instead of **kwargs to pass args in factory

Lee-000:
Tests on box2d of VOC2012Detection.

| cases               | before           | after             |
| ---------------     | ---------------- | ----------------  |
| box2d.internal_type | 59.5 µs ± 104 ns | 54.1 µs ± 99.7 ns |
| box2d.to_builtin()  | 145 µs ± 14.5 µs | 124 µs ± 292 ns   |
| box2d._data         | 90.2 µs ± 113 ns | 80.5 µs ± 3.59 µs |

### DIFF
--- a/graviti/portex/base.py
+++ b/graviti/portex/base.py
@@ -249,7 +249,7 @@ class PortexRecordBase(
 
     @property
     def _data(self) -> "ConnectedFields":  # type: ignore[override]
-        return self._fields_factory(**{name: getattr(self, name) for name in self.params})
+        return self._fields_factory({name: getattr(self, name) for name in self.params})
 
     def insert(self, index: int, name: str, portex_type: PortexType) -> None:
         """Insert the name and portex_type at the index.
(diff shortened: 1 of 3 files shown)


GravitiRob:
Task linked: [CU-2j4fdy5 Portex Factory 类型中使用字典代替 **kwargs 进行传参](https://app.clickup.com/t/2j4fdy5)

graviti/portex/base.py
```diff
@@ -249,7 +249,7 @@ class PortexRecordBase(
 
     @property
     def _data(self) -> "ConnectedFields":  # type: ignore[override]
-        return self._fields_factory(**{name: getattr(self, name) for name in self.params})
+        return self._fields_factory({name: getattr(self, name) for name in self.params})
```
linjiX, line 252: shall we use `perf` rather than `refactor`?
also shall we add perf compare result in the commit message body?

linjiX (changes requested):

PR #407, opened 2022-06-27
feat(manager): use property "parent" to replace "parent_commit_id"

graczhual:
"parent" will return a lazy loading Commit.


GravitiRob:
Task linked: [CU-2jact5g 删除 Commit.parent_commit_id 特性，使用 Commit.parent 代替，类型为 Commit](https://app.clickup.com/t/2jact5g)

graviti/manager/commit.py
```diff
@@ -72,6 +72,10 @@ def _get_commit_info(self) -> Dict[str, Any]:
                 commit_id=self.commit_id,
             )
             commit_info["committed_at"] = convert_iso_to_datetime(commit_info["committed_at"])
+            parent_commit_id = commit_info["parent_commit_id"]
```
linjiX, line 75: will it be better to use `dict.pop` instead of `dict.__getitem__` here?

linjiX (approved):

PR #414, opened 2022-06-28
feat(dataframe): add attribute "_parent" to class "Container"

graczhual:


GravitiRob:
Task linked: [CU-2jvdzqd DataFrame 添加特性 parent，拥有 parent 的 DataFrame 不允许改变长度](https://app.clickup.com/t/2jvdzqd)

graviti/dataframe/frame.py
```diff
@@ -76,6 +76,7 @@ class DataFrame(Container):
     has_keys = True
     _columns: Dict[str, Container]
     _record_key: Optional[ColumnSeries] = None
+    _parent: Optional["DataFrame"] = None
```
AChenQ, line 79: Series also has the `_parent` property, maybe we can put it into `Container`?

linjiX (changes requested):

graviti/dataframe/frame.py
```diff
@@ -388,6 +391,16 @@ def loc(self) -> DataFrameLocIndexer:
         """
         return DataFrameLocIndexer(self)
 
+    @property
+    def parent(self) -> Optional["DataFrame"]:
```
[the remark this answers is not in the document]
graczhual: Whether the users need methods to view the parent if they are not familiar with the rules.

graviti/dataframe/column/series.py
```diff
@@ -241,6 +244,11 @@ def _extend(self: _SB, values: _SB) -> None:
             5  3
 
         """
+        if self._parent is not None:
```
linjiX, line 247: we should put all checks in public `extend`, not in private `_extend`

graviti/dataframe/frame.py
```diff
@@ -198,12 +198,13 @@ def _from_pyarrow(  # type: ignore[override]
     ) -> _T:
         obj: _T = object.__new__(cls)
         obj.schema = schema
-        obj._columns = {
-            key: value.container._from_pyarrow(  # pylint: disable=protected-access
+        obj._columns = {}
+        for key, value in schema.items():
+            _value = value.container._from_pyarrow(  # pylint: disable=protected-access
```
linjiX, line 203: will it be better to set parent as a parameter of `_from_pyarrow`?

graviti/dataframe/frame.py
```diff
@@ -278,7 +280,7 @@ def _setitem(self, key: str, value: Union[Iterable[Any], _C]) -> None:
                 f"Length of values ({len(column)}) does not match length of "
                 f"self DataFrame ({len(self)})"
             )
-
+        column._parent = self  # pylint: disable=protected-access
         self.schema[key] = schema
```
linjiX, line 284: update schema first, `self.schema[key]` may raise error.

linjiX (changes requested):

graviti/dataframe/column/series.py
graczhual: Series does not support `extend` now. So we do not need to add this check because using private methods is not recommended.

linjiX (approved):

AChenQ (approved):

PR #429, opened 2022-07-07
refactor: change the max batch size to dynamic when uploading sheets

Lee-000:


graviti/operation/frame.py
```diff
@@ -82,6 +83,9 @@ class DataOperation(DataFrameOperation):  # pylint: disable=abstract-method
     def __init__(self, data: "DataFrame") -> None:
         self._data = data
 
+    def _get_max_batch_size(self) -> int:
+        return min(_MAX_BATCH_SIZE, int(_MAX_ITEMS / self._data.schema.get_column_count()))
```
linjiX, line 87: try `//`

graviti/operation/frame.py
```diff
@@ -18,6 +18,7 @@
     from graviti.dataframe import DataFrame, FileSeries
 
 _MAX_BATCH_SIZE = 2048
+_MAX_ITEMS = 60000
```
linjiX, line 21: can this limit update to `65536`?

graviti/operation/frame.py
linjiX: do we need a max value here? or should we change a larger limit?

linjiX (changes requested):

linjiX (approved):

PR #433, opened 2022-07-11
feat(portex): restrict cased characters in the "Fields" key be lowercase

Lee-000:


graviti/portex/field.py
```diff
@@ -49,7 +49,15 @@ def _check_value(cls, value: PortexType) -> None:
         if not isinstance(value, PortexType):
             raise TypeError(f'The value in "{cls.__name__}" should be a PortexType')
 
+    @classmethod
+    def _check_key(cls, key: str) -> None:
```
linjiX, line 53: what if the input key is not a string?

linjiX (changes requested):

graviti/portex/field.py
```diff
@@ -49,7 +49,18 @@ def _check_value(cls, value: PortexType) -> None:
...
+        try:
```
linjiX, line 54: do we need to call `super()._check_key` here?

linjiX (changes requested):

graviti/portex/field.py
```diff
@@ -49,7 +49,16 @@ def _check_value(cls, value: PortexType) -> None:
...
+        super()._check_key(key)
+        if not key.islower():
+            raise ValueError(
+                f'The letters in the key of class {cls.__name__}, "{key}", must be in lowercase'
+            )
+
     def _setitem(self, key: str, value: PortexType) -> None:
+        self._check_key(key)
         self._check_value(value)
         super()._setitem(key, value)
```
linjiX, line 63: will `super()._setitem` calls `_check_key`?

linjiX (changes requested):

linjiX (changes requested):

graviti/portex/field.py
```diff
@@ -49,6 +49,14 @@ def _check_value(cls, value: PortexType) -> None:
         if not isinstance(value, PortexType):
             raise TypeError(f'The value in "{cls.__name__}" should be a PortexType')
 
+    @classmethod
+    def _check_key(cls, key: str) -> None:
+        super()._check_key(key)
+        if not key.islower():
+            raise ValueError(
+                f'The letters in the key of class {cls.__name__}, "{key}", must be in lowercase'
```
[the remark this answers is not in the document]
linjiX: ref the docstring of `str.islower`

graviti/portex/field.py
```diff
@@ -49,6 +50,14 @@ def _check_value(cls, value: PortexType) -> None:
         if not isinstance(value, PortexType):
             raise TypeError(f'The value in "{cls.__name__}" should be a PortexType')
 
+    @classmethod
+    def _check_key(cls, key: str) -> None:
+        super()._check_key(key)
+        if re.search("[A-Z]", key):
```
linjiX, line 56: can we use `not str.isupper` rather than regex?

linjiX (changes requested):

graviti/portex/field.py
Lee-000: `not str.isupper` is True when not all letters in the string are uppercase. It does not distinguish cases like `abC` and `abc`.

linjiX (approved):

PR #444, opened 2022-07-13
feat(dataframe): implement "__setitem__" of loc&iloc in column series

AChenQ:


graviti/dataframe/column/indexing.py
```diff
@@ -43,9 +43,9 @@ def __setitem__(self, key: int, value: Any) -> None:
     def __setitem__(
         self,
         key: Union[int, slice],
-        value: Union[Iterable[Any], Any],
+        value: Union[Any, Iterable[Any], "SeriesBase"],
     ) -> None:
-        pass
+        self.obj.__setitem__(key, value)
```
linjiX, line 48: `obj.__setitem__` is a columnar operation, but `loc.__setitem__` is a row operation?

PR #447, opened 2022-07-14
fix(dataframe): split "_to_request_data" method to fix None record error

AChenQ:


graviti/operation/frame.py
```diff
@@ -272,7 +270,7 @@ def do(  # pylint: disable=invalid-name
             file_pbar: The process bar for uploading binary files.
 
         """
-        data = self._data._to_request_data()  # pylint: disable=protected-access
+        data = self._data._to_post_data()  # pylint: disable=protected-access
```
graczhual, line 273: _to_patch_data?

graczhual (approved):

linjiX (approved):

PR #448, opened 2022-07-14
feat(utility): add FileBase, File and RemoteFile for basic binary files

Lee-000:


GravitiRob:
Task linked: [CU-2mez3rv 实现基础 File 类型，支持提取文件后缀 + 文件大小](https://app.clickup.com/t/2mez3rv)

graviti/utility/binary_file.py
```diff
@@ -0,0 +1,243 @@
...
+        """Get the key of the file.
+
+        Returns:
+            The key of the file.
+
+        """
+        return self._key
+
+    @property
+    def format(self) -> str:
```
linjiX, line 55: should format be `Optional`

graviti/utility/binary_file.py
```diff
@@ -0,0 +1,243 @@
...
+
+        Returns:
+            The File or RemoteFile instance.
+
+        """
+        key = scalar["key"].as_py()
+        format_: str = scalar["format"].as_py()
+        size: int = scalar["size"].as_py()
+
+        if key.startswith("file://"):
```
linjiX, line 90: the key of remote file is not starts with `oss://` or `s3://`, is it a good choice to add `file://` suffix for local file?

linjiX (changes requested):

graviti/utility/binary_file.py
Lee-000: But we need to distinguish the local file and the remote file with the key.

graviti/utility/binary_file.py
Lee-000: The format could be an empty string.

linjiX: None or empty string, the schema says the format is nullable?

graviti/utility/binary_file.py
linjiX: we need a better way to distinguish the local and remote file.

graviti/utility/binary_file.py
```diff
@@ -0,0 +1,214 @@
...
+    def size(self) -> int:
+        """Get the size of the file.
+
+        Returns:
+            The size of the file.
+
+        """
+        return self._size
+
+    def to_post_data(self) -> Dict[str, Union[int, str]]:
```
linjiX, line 64: should this be private?

graviti/utility/binary_file.py
```diff
@@ -0,0 +1,214 @@
...
+    @property
+    def key(self) -> str:
+        """Get the key of the file.
+
+        Returns:
+            The key of the file.
+
+        """
+        if not hasattr(self, "_key"):
+            self._key = self._path.as_uri()
```
linjiX, line 122: why the key is `uri`?

graviti/utility/binary_file.py
```diff
@@ -0,0 +1,214 @@
...
+
+        Returns:
+            The extension of the file.
+
+        """
+        if not hasattr(self, "_extension"):
+            suffix = self._path.suffix
+            if suffix in _ENCODINGS:
+                suffix = Path(self._path.stem).suffix + suffix
+            self._extension = suffix.lstrip(".")
```
linjiX, line 137: will it be better to keep the leading `.`

linjiX (changes requested):

graviti/utility/binary_file.py
```diff
@@ -0,0 +1,208 @@
...
+    @property
+    def key(self) -> str:
+        """Get the key of the file.
+
+        Returns:
+            The key of the file.
+
+        """
+        if not hasattr(self, "_key"):
+            self._key = self._path.as_posix()
```
linjiX, line 116: why using `as_posix`?

linjiX (approved):

PR #457, opened 2022-07-19
docs: remove the limitation of enum values in "schema"

AChenQ:


docs/source/features/sheet_management/schema.rst
```diff
@@ -95,9 +95,6 @@ An enum type can be created by giving the values:
    >>> enum.values
    ['a', 'b', 'c']
 
-.. note::
```
graczhual, line 98: do we need to add description about enum schema like we need to use int value in data?

AChenQ: Users dont't need to use int value in data.

PR #475, opened 2022-07-21
fix(paging): correct the return type of "get_slice" in mapped page

linjiX:


GravitiRob:
Task linked: [CU-2n00czn 实现带 mapper 的 Page 使用 mapper 代替 convertor](https://app.clickup.com/t/2n00czn)

PR #482, opened 2022-07-25
refactor(file): move binary file related classes to file module

Lee-000:


graviti/file/__init__.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python3
+#
+# Copyright 2022 Graviti. Licensed under MIT License.
+#
+"""File module."""
+
+from graviti.file.base import File, RemoteFile
```
linjiX, line 7: set yourself as the owner of this module in CODEOWNERS?

linjiX (approved):

PR #487, opened 2022-07-28
feat(dataframe): forbid deleting items by loc for non-root Containers

AChenQ:
Not support '__delitem__' for the Series which is a member of a
DataFrame.

Not support 'loc.__delitem__' and 'iloc.__delitem__' for the DataFrame
which is a member of another DataFrame

### DIFF
--- a/graviti/dataframe/column/series.py
+++ b/graviti/dataframe/column/series.py
@@ -115,7 +115,12 @@ class SeriesBase(Container):  # pylint: disable=abstract-method
         return self._data[key]
 
     def __delitem__(self, key: Union[int, slice]) -> None:
-        self._data.__delitem__(key)
+        if self._root is not None:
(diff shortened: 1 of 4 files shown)


graviti/dataframe/column/series.py
```diff
@@ -115,7 +115,12 @@ def __getitem__(self, key: int) -> Any:
         return self._data[key]
 
     def __delitem__(self, key: Union[int, slice]) -> None:
-        self._data.__delitem__(key)
+        if self._root is not None:
```
linjiX, line 118: ref commit message:
`feat(dataframe): forbid deleting items by loc for non-root Containers`

linjiX (changes requested):

linjiX (approved):

PR #504, opened 2022-08-12
feat(operation): update local files to remote files before uploading

graczhual:


GravitiRob:
Task linked: [CU-2p2hdke 上传数据相关 operation 中上传二进制文件后更新 request_data](https://app.clickup.com/t/2p2hdke)

graviti/operation/frame.py
```diff
@@ -89,6 +94,39 @@ def _get_max_batch_size(self) -> int:
...
+    def _update_local_files(
+        self,
+        object_policy_manager: "ObjectPolicyManager",
+    ) -> List[List[Optional[Tuple[str, Path]]]]:
+        # object_policy_manager = self._data.object_policy_manager
+        if not object_policy_manager:
+            raise ValueError("Require object policy manager to upload data")
+
+        # pylint: disable=protected-access
+        prefix = object_policy_manager._init_put_policy()["policy"][  # type: ignore[attr-defined]
```
linjiX, line 106: shall we add a method or property to policy manager for getting the `prefix`?

graviti/operation/frame.py
```diff
@@ -89,6 +94,39 @@ def _get_max_batch_size(self) -> int:
...
+            ]
+            key_and_path: List[Optional[Tuple[str, Path]]] = []
+            for index, file in enumerate(file_array._data):
+                if isinstance(file, File):
+                    key = f"{prefix}/{file.get_checksum()}"
+                    key_and_path.append((key, file.path))
+
+                    post_data = file._to_post_data()
+                    post_data["key"] = key
+                    file_array[index] = file_type(**post_data)  # type: ignore[arg-type]
```
linjiX, line 124: the file_array comes from a copied DataFrame, update file_array will not effect the sheet.

graviti/operation/frame.py
```diff
@@ -351,3 +385,33 @@ def do(  # pylint: disable=invalid-name, unused-argument
             sheet=sheet,
             record_keys=self.record_keys,
         )
+
+
+def _upload_files(
+    keys_and_paths: Iterable[Tuple[str, Path]],
```
linjiX, line 391: why we need `keys_and_paths`, can use just use File list instead?

linjiX (changes requested):

graviti/operation/frame.py
graczhual: we need `keys_and_paths` to determine whether the data is remote or local. Because the data in operation has been changed to remote to get the True post data.

linjiX: class `File` means local file and `RemoteFile` means remote file. why not use the class to determine whether the data is remote of local?

PR #507, opened 2022-08-15
feat: add property "object_policy_manager" to class "LazyFactoryBase"

graczhual:


graviti/manager/sheets.py
```diff
@@ -84,11 +84,11 @@ def _init_dataframe(self, sheet_name: str) -> DataFrame:
             partial(self._list_data, sheet_name=sheet_name),
             pa.struct([pa.field(RECORD_KEY, pa.string()), *patype]),
         )
-
-        dataframe = DataFrame._from_factory(  # pylint: disable=protected-access
-            factory, schema, self._dataset.object_policy_manager
-        )
+        # pylint: disable=protected-access
+        factory._object_policy_manager = self._dataset.object_policy_manager
```
linjiX, line 88: do we need a property setter to get rid of the pylint warning?

linjiX (approved):

graviti/manager/sheets.py
graczhual: The setter has been added, but I'm not sure which one is better.

linjiX (approved):

PR #508, opened 2022-08-16
feat(dataframe): refresh the data of dataframe after uploading

graczhual:


graviti/manager/sheets.py
```diff
@@ -169,17 +169,30 @@ def _upload_to_draft(self, draft_number: int, jobs: int, quiet: bool) -> None:
...
-                            dataset.url,
-                            dataset.owner,
-                            dataset.name,
-                            draft_number=draft_number,
-                            sheet=sheet_name,
-                            jobs=jobs,
-                            data_pbar=data_pbar,
-                            file_pbar=file_pbar,
-                            object_policy_manager=object_policy_manager,
+                    if dataframe.operations:
```
linjiX, line 172:
```python
if not dataframe.operations:
    continue
```

linjiX (changes requested):

graviti/dataframe/column/series.py
```diff
@@ -208,6 +208,24 @@ def _get_repr_indices(self) -> Iterable[int]:
     def _del_item_by_location(self, key: Union[int, slice]) -> None:
         self._data.__delitem__(key)
 
+    @classmethod
```
linjiX, line 211: maybe remove the "support" in the commit message is better.
```
feat(dataframe): refresh the data of dataframe after uploading
```
`support` sounds like the the new feature is optional, but here this refresh feature is not an optional feature.

linjiX (changes requested):

linjiX (approved):

PR #526, opened 2022-08-30
fix(manager): fix dataset.commit failed when the sheet list is unchanged

graczhual:


graviti/manager/dataset.py
```diff
@@ -249,7 +249,13 @@ def commit(
                 "Please checkout a branch first"
             )
         if not head.operations:
-            raise StatusError("It is not allowed to commit a dataset without any modifications")
+            without_modifications = True
+            for dataframe in head.values():
```
linjiX, line 253: try:
```python
modified = head.operations or any(df.operations for df in head.values())
```

linjiX (changes requested):

graviti/manager/dataset.py
```diff
@@ -249,7 +249,13 @@ def commit(
                 "Please checkout a branch first"
             )
         if not head.operations:
-            raise StatusError("It is not allowed to commit a dataset without any modifications")
+            without_modifications = True
```
linjiX, line 252: is it possible to describe the issue more detailed in the commit message:
here is a reference:
```
fix(manager): fix dataset.commit failed when the sheet list is unchanged
```

linjiX (changes requested):

linjiX (approved):

